PR #17, opened 2022-08-19
Fix integer overflow in IsIoReadWriteAllowed()

### LINKED ISSUE #6: Potential integer overflow in IsIoReadWriteAllowed()
The supervisor calls IsIoReadWriteAllowed() to see if a user can read or write to specific io ports. The code looks pretty decent; however, a few minor things seem wrong. There's some weirdness around casting and truncating when calling it (and when later using the ioports), which looks kinda bad, but ultimately it doesn't seem to cause much problems.

When validating the ioport and size, there is an integer overflow. It will overflow if the port is really big (0xfffffffd or larger). If there is a policy to allow writes to low ports (e.g., 0), then it could allow writing to a high port (e.g., 65535, 0xffffffff truncated to 16 bits). Ultimately this is a really specific set of circumstances and probably not that big of a deal, but it does look like a clear bug that should be addressed.

Fix recommendation:
A safe int function to convert port (IoAddress) to prevent any integer overflow from occurring. Also added an integer overflow check as a defense in depth mechanism.

Acknowledgement:

kuqin12:
Bug Description:
The supervisor calls IsIoReadWriteAllowed() to see if a user can read or write to specific io ports. The code looks pretty decent; however, a few minor things seem wrong. There's some weirdness around casting and truncating when calling it (and when later using the ioports), which looks kinda bad, but ultimately it doesn't seem to cause many problems.

When validating the ioport and size, there is an integer overflow. It will overflow if the port is really big (0xfffffffd or larger). If there is a policy to allow writes to low ports (e.g., 0), then it could allow writing to a high port (e.g., 65535, 0xffffffff truncated to 16 bits). Ultimately this is a really specific set of circumstances and probably not that big of a deal, but it does look like a clear bug that should be addressed.

Fix:
A safe int function to convert port (IoAddress) to prevent any integer overflow from occurring. Also added an integer overflow check as a defense in depth mechanism.

fixes #6

### DIFF
--- a/MmSupervisorPkg/Library/SmmPolicyGateLib/SmmPolicyGateLib.c
+++ b/MmSupervisorPkg/Library/SmmPolicyGateLib/SmmPolicyGateLib.c
@@ -15,6 +15,7 @@
 #include <Library/DebugLib.h>
 #include <Library/SmmPolicyGateLib.h>
 #include <Library/SysCallLib.h>
+#include <Library/SafeIntLib.h>
 
 /**
   Given an IO port address and size, determine if the request is allowed by
@@ -50,6 +51,7 @@ IsIoReadWriteAllowed (
   UINT32                                     IoSize        = 0;
   UINT32                                     i;
   BOOLEAN                                    FoundMatch = FALSE;
+  UINT16                                     Dummy;
 
   //
   // Check to ensure that only one of SECURE_POLICY_RESOURCE_ATTR_READ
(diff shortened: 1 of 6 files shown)


MmSupervisorPkg/Library/SmmPolicyGateLib/SmmPolicyGateLib.c
```diff
@@ -78,6 +80,27 @@ IsIoReadWriteAllowed (
...
+    Status = EFI_INVALID_PARAMETER;
+    goto Exit;
+  }
+
+  //
+  // Check to ensure the ending IO port number is within the range of 16bit value
+  // The MAX_UINT16 + 1 is because access 1 byte at 0xFFFF is still legit, if needed
+  //
+  if (IoAddress + IoSize > MAX_UINT16 + 1) {
+    // IO port 0xFFFF with width of
```
spbrogan, line 98: incomplete comment

spbrogan (approved):

MmSupervisorPkg/Library/SmmPolicyGateLib/SmmPolicyGateLib.c
makubacki, line 98: Looks like this comment is incomplete

makubacki (approved):

MmSupervisorPkg/Library/SmmPolicyGateLib/SmmPolicyGateLib.c
makubacki: I see Sean also mentioned this. I'll close this comment.

kuqin12: This is removed.

PR #20, opened 2022-08-19
Syscall entry/exit register save / restore (general purpose and FPU/MMX/SSE/AVX)

### LINKED ISSUE #9: Syscall entry/exit register save / restore (general purpose and FPU/MMX/SSE/AVX)
SyscallCenter implemented in SysCallEntry.nasm is called as the supervisor entry point when user space issues a SYSCALL instruction. This function is responsible for proper setup, calling the dispatcher, and then returning to user space properly.

Except, it doesn't seem to restore all registers. It restores a subset of the general-purpose registers (r11, rbx, rdi, r12, rsi, r9, r8, rdx, rbp rcx and rsp), clearly leaving out r10, r13, r14, and r15. If the dispatcher uses any of these registers, they would get returned to user space.

In addition to the general-purpose registers, it doesn't seem to restore any of the FPU/MMX/SSE/AVX registers. Optimizing compilers tend to use the SIMD registers and instructions for things like memcpy().

Fix Recommendation: 
Consider saving/restoring all general-purpose registers.

Acknowledgement:

kuqin12:
Bug Description:
SyscallCenter implemented in SysCallEntry.nasm is called as the supervisor entry point when user space issues a SYSCALL instruction. This function is responsible for proper setup, calling the dispatcher, and then returning to user space properly.

Except, it doesn't seem to restore all registers. It restores a subset of the general purpose registers (r11, rbx, rdi, r12, rsi, r9, r8, rdx, rbp rcx and rsp), clearly leaving out r10, r13, r14, and r15. If the dispatcher uses any of these registers, they would get returned to user space.

In addition to the general purpose registers, it doesn't seem to restore any of the FPU/MMX/SSE/AVX registers. Optimizing compilers tend to use the SIMD registers and instructions for things like memcpy().

Fix:
In the syscall interface:
- Adding saving/restoring all general purpose registers.
- Saving/restoring MMX/SSE/FPU registers with FXSAVE and FXSTOR.
- AVX operations are prohibited by not enabling OSXSAVE in CR4.

The instructions addition is measured to have <100ms slowdown for 1M syscall calls.

fixes #9

### DIFF
--- a/MmSupervisorPkg/Core/PrivilegeMgmt/SysCallEntry.nasm
+++ b/MmSupervisorPkg/Core/PrivilegeMgmt/SysCallEntry.nasm
@@ -73,6 +73,15 @@ ASM_PFX(SyscallCenter):
     push    rdi
     push    rbx
     push    r11
+    push    r10
+    push    r13
+    push    r14
+    push    r15
+
+    ;; FX_SAVE_STATE_X64 FxSaveState;
+    sub rsp, 512
+    mov rdi, rsp
+    db 0xf, 0xae, 0x7 ;fxsave [rdi]
 
     ;Prepare for ds, es, fs, gs
     xor     rbx, rbx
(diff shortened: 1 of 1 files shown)


spbrogan (approved):

makubacki (approved):

kuqin12:
A performance test was conducted with this change on both physical FASR systems.

Without the change:
- 100k rounds of MSR request took 166ms
- 100k rounds of IO 8bit request took 5693ms

With the change:
- 100k rounds of MSR request took 219ms
- 100k rounds of IO 8bit request took 5813ms

A simple performance test tool can be found here: https://github.com/kuqin12/mu_feature_mm_supv/tree/supv_fix_9_reg_perf

PR #24, opened 2022-08-20
Fixed GUID comparison that could lead to arbitrary write from FreePool()

### LINKED ISSUE #13: Race condition in SmiHandlerProfileUnregisterHandler() that could lead to FreePool() of something that isn't pool allocated
SmiHandlerProfileUnregisterHandler() is called by ProcessUserHandlerUnreg() which is called by the supervisor when handling SMM_MM_HDL_UNREG_1/SMM_MM_HDL_UNREG_2 syscalls. It takes a guid pointer (as well as a context and context size) provided by user space. It has been inspected with a call to InspectTargetRangeOwnership() to ensure it belongs to user space. SmiHandlerProfileUnregisterHandler() uses the guid pointer twice. Once to look up the handler (and its associated context) and a second time to free memory if the guid content is the same as gEfiSmmUsbDispatch2ProtocolGuid.

Given that the guid points to user space controlled data, it is possible that this content could have changed between the first time it was used and the second time it was used. I'm not sure if user space gets access to more than one core (in which case they could instruct the 2nd core to modify the data asynchronously). If the guid content is not equal to gEfiSmmUsbDispatch2ProtocolGuid during the first use but is equal during the second use, user space can induce the supervisor to FreePool() data that user space controls instead of supervisor allocated pool memory.

In addition, MmInstallProtocolInterfaceNotify() only has an assert (this is reachable through the SMM_INST_PROT syscall with a similar user-provided guid pointer) that could be triggered with a similar guid race where multiple entries in a list could have the same guid. But this is benign because it occurs after initial MM launching and does not impact supervisor behavior.

One last note, SmiHandlerProfileUnregisterHandler() has a second case that compares against a specific guid (gEfiSmmSwDispatch2ProtocolGuid) and returns a supervisor pool allocated piece of memory. However, in this case, there is no logic to free this memory at the end of the function, leading to a memory leak. In and on itself, memory leaks aren't great, but (certainly in the context of UEFI/SMM) I wouldn't call them out as a security issue. However, exploitation of the "MmInstallConfigurationTable()" issue described above would likely benefit from a reliable memory leak, and as such, this might have some security consequences.

Fix Recommendation:
- Consider capturing the user-provided guid before using it. 
- Fix the memory leak. 

Acknowledgement:

kuqin12:
Bug Description:
SmiHandlerProfileUnregisterHandler() is called by ProcessUserHandlerUnreg() which is called by the supervisor when handling SMM_MM_HDL_UNREG_1/SMM_MM_HDL_UNREG_2 syscalls. It takes a guid pointer (as well as a context and context size) provided by user space. It has been inspected with a call to InspectTargetRangeOwnership() to ensure it belongs to user space. SmiHandlerProfileUnregisterHandler() uses the guid pointer twice. Once to look up the handler (and its associated context) and a second time to free memory if the guid content is the same as gEfiSmmUsbDispatch2ProtocolGuid.

Given that the guid points to user space controlled data, it is possible that this content could have changed between the first time it was used and the second time it was used. I'm not sure if user space gets access to more than one core (in which case they could instruct the 2nd core to modify the data asynchronously). If the guid content is not equal to gEfiSmmUsbDispatch2ProtocolGuid during the first use but is equal during the second use, user space can induce the supervisor to FreePool() data that user space controls instead of supervisor allocated pool memory.

In addition, MmInstallProtocolInterfaceNotify() only has an assert (this is reachable through the SMM_INST_PROT syscall with a similar user-provided guid pointer) that could be triggered with a similar guid race where multiple entries in a list could have the same guid. But this is benign because it occurs after initial MM launching and does not impact supervisor behavior.

One last note, SmiHandlerProfileUnregisterHandler() has a second case that compares against a specific guid (gEfiSmmSwDispatch2ProtocolGuid) and returns a supervisor pool allocated piece of memory. However, in this case, there is no logic to free this memory at the end of the function, leading to a memory leak. In and on itself, memory leaks aren't great, but (certainly in the context of UEFI/SMM) I wouldn't call them out as a security issue. However, exploitation of the "MmInstallConfigurationTable()" issue described above would likely benefit from a reliable memory leak, and as such, this might have some security consequences.

The impact of handler profile related bugs are trivial because this is a test/development feature, which should not be enabled in the release firmware.

Fix:
- Capturing the user-provided GUID before using it.
- Fixed the memory leak.

fixes #13

### DIFF
--- a/MmSupervisorPkg/Core/Hand/Handle.c
+++ b/MmSupervisorPkg/Core/Hand/Handle.c
@@ -219,6 +219,7 @@ MmInstallProtocolInterfaceNotify (
   IHANDLE             *Handle;
   EFI_STATUS          Status;
   VOID                *ExistingInterface;
+  EFI_GUID            TempProtocolGuid;
 
   //
   // returns EFI_INVALID_PARAMETER if InterfaceType is invalid.
@@ -233,16 +234,21 @@ MmInstallProtocolInterfaceNotify (
   }
 
   //
+  // Capture the incoming GUID. This is necessary for mitigating TOCTOU issue.
+  //
+  CopyMem (&TempProtocolGuid, Protocol, sizeof (EFI_GUID));
+
+  //
   // Print debug message
   //
-  DEBUG ((DEBUG_LOAD | DEBUG_INFO, "MmInstallProtocolInterface: %g %p\n", Protocol, Interface));
+  DEBUG ((DEBUG_LOAD | DEBUG_INFO, "MmInstallProtocolInterface: %g %p\n", &TempProtocolGuid, Interface));
 
   Status = EFI_OUT_OF_RESOURCES;
   Prot   = NULL;
   Handle = NULL;
 
   if (*UserHandle != NULL) {
-    Status = MmHandleProtocol (*UserHandle, Protocol, (VOID **)&ExistingInterface);
+    Status = MmHandleProtocol (*UserHandle, &TempProtocolGuid, (VOID **)&ExistingInterface);
     if (!EFI_ERROR (Status)) {
       return EFI_INVALID_PARAMETER;
     }
@@ -251,7 +257,7 @@ MmInstallProtocolInterfaceNotify (
   //
   // Lookup the Protocol Entry for the requested protocol
   //
-  ProtEntry = MmFindProtocolEntry (Protocol, TRUE);
+  ProtEntry = MmFindProtocolEntry (&TempProtocolGuid, TRUE);
   if (ProtEntry == NULL) {
     goto Done;
   }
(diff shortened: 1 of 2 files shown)


spbrogan (approved):

makubacki (approved):

makubacki:
As a a general comment, you might want to preface your commit descriptions i these PRs with something like "Bug Description: <text>". The context is a little strange because people would expect the fix described rather than the issue describe. Later "Fix:" helps connect what it is but that is kind of far down. I personally think you could minimally describe the actual code and let the link to the issue describe the issue.

spbrogan (approved):

PR #35, opened 2022-10-10
Update package and module version

kuqin12:
This change bumps the version number to make a release before we transition to 202208.


makubacki:
@kuqin12, I assume you're planning to bump the major version after 202208?

kuqin12:
> @kuqin12, I assume you're planning to bump the major version after 202208?

yep. 6.002 and 6.003 only differs minimally. We could bump the major version up to 7 once we are 202208 based.

kenlautner (approved):

PR #39, opened 2022-10-21
[Rebase & FF] Fixing issues discovered during GCC run

kuqin12:
GCC compiled supervisor is being exercised on the MU_TIANO_PLATFORMS repo, which exposed a few cases not verified during development process. These changes are made to fix them.


makubacki (approved):

MmSupervisorPkg/Core/MmSupervisorCore.inf
```diff
@@ -22,16 +22,16 @@
   BASE_NAME                      = MmSupervisorCore
   FILE_GUID                      = 4E4C89DC-A452-4B6B-B183-F16A2A223733
   MODULE_TYPE                    = MM_CORE_STANDALONE
-  VERSION_STRING                 = 6.003
+  VERSION_STRING                 = 7.001
```
kenlautner, line 25: Just curios but what is the criteria for updating the major version of the core?  Just want to know why it's being update here versus other times I've seen the core updated.

MmSupervisorPkg/Drivers/MmPeiLaunchers/Common/MmIplCommon.c
```diff
@@ -397,8 +397,12 @@ QuerySupervisorVersion (
     ));
 
   if (VersionInfo->Version == 0) {
+ #ifdef __GNUC__
```
kenlautner, line 400: Is it correct that we want this function to return EFI_SUCCESS if we are using the GCC combiler?  I'd assume we'd want it to still return EFI_SECURITY_VIOLATION.

MmSupervisorPkg/Core/MmSupervisorCore.inf
kuqin12: The version will be bumped up whenever it is ready to make a release :) This change update the major version because the MU foundation is updated to 202208. I meant to do that after 202208 support is in, but it caused some issues when building with GCC. These should fix those issues and make it ready enough for a release.

MmSupervisorPkg/Drivers/MmPeiLaunchers/Common/MmIplCommon.c
kuqin12: That will make GCC built supervisor never boot (which is part of the GCC building issues.) There is currently no effortless way to embed these versions into the final output.

At least for now, we would not check this on GCC built systems. But if others ever saw this message and have better ideas, please comment and welcome to contribute directly.

kenlautner (approved):

MmSupervisorPkg/Drivers/MmPeiLaunchers/Common/MmIplCommon.c
kenlautner: Got it.  I assumed this was intended but just wanted to make sure.

kuqin12: Right. This is only a workaround, so that it does not break the build all the time. But we will keep investigating with other methods, hoping to remove this branching logic some time.

PR #41, opened 2022-11-05
[Rebase & FF] Add stack cookie

kuqin12:
## Description

This change added the support for stack cookie in standalone MM core and user modules:
- Added assembly function to write to `ICC_SGI1R` from non-secure EL1/EL2
- Extended current `ArmGicSendSgiTo` interface to also support sending SGI on GIC v3 and v4
- Added a new PSCI command index to query PSCI suspend features.

The assembly support is intentionally left out as the common solution for MSVC and GCC
support is still under development.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on both GCC and MSVC compiled Q35 platform.

## Integration Instructions

Need to include the appropriate instance of `BaseBinSecurityLib` in the platform dsc file to pass the build.

### DIFF
--- a/MmSupervisorPkg/Core/Dispatcher/Dispatcher.c
+++ b/MmSupervisorPkg/Core/Dispatcher/Dispatcher.c
@@ -292,8 +292,10 @@ MmLoadImage (
 {
   UINTN                         PageCount;
   EFI_STATUS                    Status;
+  EFI_STATUS                    StackCookieStatus;
   EFI_PHYSICAL_ADDRESS          DstBuffer;
   PE_COFF_LOADER_IMAGE_CONTEXT  ImageContext;
+  UINT64                        *SecurityCookieAddress;
 
   DEBUG ((DEBUG_INFO, "MmLoadImage - %g\n", &DriverEntry->FileName));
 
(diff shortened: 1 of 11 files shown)


makubacki:
@kuqin12, is this ready for review?

.pytool/CISettings.py
```diff
@@ -124,6 +124,7 @@ def GetName(self):
     def GetPackagesPath(self):
         ''' Return a list of workspace relative paths that should be mapped as edk2 PackagesPath '''
         result = [
+            self.GetWorkspaceRoot (),
```
kuqin12, line 127: This should not be here

MmSupervisorPkg/Core/Dispatcher/Dispatcher.c
```diff
@@ -292,9 +292,12 @@ MmLoadImage (
 {
   UINTN                         PageCount;
   EFI_STATUS                    Status;
+  EFI_STATUS                    StackCookieStatus;
   EFI_PHYSICAL_ADDRESS          DstBuffer;
   PE_COFF_LOADER_IMAGE_CONTEXT  ImageContext;
 
+  UINT64  *SecurityCookieAddress;                           // MS_CHANGE_?
```
kuqin12, line 299: Not really ms_change anymore

kuqin12:
Also need a doc change to indicate the vector number.

.pytool/CISettings.py
kuqin12: Removed. See if this will complete building as expected.

MmSupervisorPkg/Core/Dispatcher/Dispatcher.c
kuqin12: Removed.

apop5 (approved):

PR #43, opened 2022-11-23
Updated dependencies to latest

kenlautner:
Updated dependencies to be top of tree so that platforms relying on this repo don't run into tracking errors.


TaylorBeebe (approved):

makubacki:
You might want to mention this is a breaking change for platforms that do not update to the commits referenced in those repos here.

makubacki (approved):

PR #57, opened 2023-01-18
Converge demotion routines in assembly into one and update the stack alignments

kuqin12:
# Preface

Please ensure you have read the [contribution docs](https://github.com/microsoft/mu/blob/master/CONTRIBUTING.md) prior
to submitting the pull request. In particular,
[pull request guidelines](https://github.com/microsoft/mu/blob/master/CONTRIBUTING.md#pull-request-best-practices).

## Description

In the original implementation, each operation privilege transitioning has its own function, which is mostly alike, causing the same change having to be replicated four times each change involving these routines..

This change mainly converges the 4 routines into a general-purpose assembly routine `InvokeDemotedRoutine`, which accepts variable input arguments and transition them into desired function, for both BSP and AP procedures.

In the meantime, this change also fixes stack alignment discrepancy when the program is built on GCC vs. VC by caching the input stack in RBP. This should fix the issue reported: https://github.com/microsoft/mu_tiano_platforms/issues/111.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change was tested on QEMU Q35 platform built with both GCC and VC, as well as on proprietary hardware. All binaries can boot to Windows OS.

## Integration Instructions

N/A

### DIFF
--- a/MmSupervisorPkg/Core/PrivilegeMgmt/CallGateTransfer.c
+++ b/MmSupervisorPkg/Core/PrivilegeMgmt/CallGateTransfer.c
@@ -213,3 +213,23 @@ CallgateInit (
       );
   }
 }
+
+/**
+  Invoke MM driver in CPL 3.
+**/
+EFI_STATUS
+EFIAPI
+InvokeDemotedDriverEntryPoint (
+  IN MM_IMAGE_ENTRY_POINT  *EntryPoint,
+  IN EFI_HANDLE            ImageHandle,
+  IN EFI_MM_SYSTEM_TABLE   *MmSystemTable
+  )
+{
+  return InvokeDemotedRoutine (
+          mSmmMpSyncData->BspIndex,
+          (EFI_PHYSICAL_ADDRESS)(UINTN)EntryPoint,
+          2,
+          ImageHandle,
+          MmSystemTable
+        );
+}
--- a/MmSupervisorPkg/Core/PrivilegeMgmt/CallGateTransfer.c
+++ b/MmSupervisorPkg/Core/PrivilegeMgmt/CallGateTransfer.c
@@ -228,12 +228,12 @@ InvokeDemotedDriverEntryPoint (
   }
 
   return InvokeDemotedRoutine (
-          mSmmMpSyncData->BspIndex,
-          (EFI_PHYSICAL_ADDRESS)(UINTN)EntryPoint,
-          2,
-          ImageHandle,
-          MmSystemTable
-        );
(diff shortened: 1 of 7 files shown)


MmSupervisorPkg/Core/PrivilegeMgmt/PrivilegeMgmt.h
```diff
@@ -118,6 +106,18 @@ SyncMmEntryContextToCpl3 (
   VOID
   );
 
+/**
+  Invoke specified routine on specified core in CPL 3.
+**/
```
makubacki, line 111: For this function in particular, I'd provide argument descriptions since manually providing a variable argument count is a bit unusual (but understandable in this case).

MmSupervisorPkg/Core/PrivilegeMgmt/CallGateTransfer.c
```diff
@@ -195,21 +195,132 @@ CallgateInit (
...
+
+/**
+  Invoke MM driver in CPL 3.
+**/
+EFI_STATUS
+EFIAPI
+InvokeDemotedDriverEntryPoint (
+  IN MM_IMAGE_ENTRY_POINT  *EntryPoint,
+  IN EFI_HANDLE            ImageHandle,
+  IN EFI_MM_SYSTEM_TABLE   *MmSystemTable
```
makubacki, line 223: Are there cases where this function is invoked and `MmSystemTable` is `NULL`?

makubacki (approved):

MmSupervisorPkg/Core/PrivilegeMgmt/CallGateTransfer.c
kuqin12: Yes, the first driver (ideally `Ring3Broker`) loaded will have a NULL pointer for this field. This is because the table initiated in MM core is useless for the user drivers directly due to privilege mismatch.

The user version of system table is registered by the `Ring3Broker` once it is loaded and all the subsequent module entrypoints will be loaded with the user version of `MmSystemTable`.

MmSupervisorPkg/Core/PrivilegeMgmt/PrivilegeMgmt.h
kuqin12: Thanks, updated.

mikeytdisco (approved):
Tested on my QemuQ35Pkg test platform.

PR #101, opened 2023-03-29
Untangle syscall entry logging using feature PCD

### LINKED ISSUE #99: [Feature]: Control Syscall Enter/Exit Message Separately from DEBUG_VERBOSE
### Feature Overview

The `SysCallDispatcher()` function in [MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c](https://github.com/microsoft/mu_feature_mm_supv/blob/8b8aeef82a98410e61a632f2eaf41dd161087501/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c#L245-L256) prints a message upon every entry and exit to the function.

https://github.com/microsoft/mu_feature_mm_supv/blob/8b8aeef82a98410e61a632f2eaf41dd161087501/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c#L245-L256

The print error level is `DEBUG_VERBOSE` but the message frequency greatly exceeds the number of messages typically printed in a single function at that level. It effectively makes a boot with `DEBUG_VERBOSE` level set impossible to use in a reasonable developer cycle.

On one system, it has printed 86,000+ times on a single boot (before being powered off). On the UART at 115200, this has consumed 39+ minutes and the text dominates the serial log.

This is useful information, but it's not general verbose information. It's extreme detail for a very specific scenario that I believe is best handled with a special flag.

This request does not specify how this flag is implemented.

> Note: This is submitted as a feature request, but could also be considered a bug.

### Solution Overview

A specific solution is not given. Options could include:

- Define a print error level for "MM Supervisors" (or something more relevant) in `gEfiMdePkgTokenSpaceGuid.PcdFixedDebugPrintErrorLevel`

- Define a print error level PCD in 'MmSupervisorPkg` and add a bit for the sys call dispatcher

- Define another mechanism like a dedicated (e.g. boolean) PCD to control whether this message is printed

### Alternatives Considered

1. Override the `PrintErrorLevel` on the MmSupervisorCore module to remove `DEBUG_VERBOSE`

Not desirable as other verbose messages in the module are missed.

2. Comment out this message.

Not desirable since source code modification should not be necessary for this specific case.

### Urgency

Medium

Someone else needs to implement the feature

Maintainer feedback requested

### Anything else?

_No response_

kuqin12:
## Description

Resolves #99 

This change will introduce a new PCD to untangle the syscall entry logging from verbose level. Per feedback from this issue: https://github.com/microsoft/mu_feature_mm_supv/issues/99, the logging is too intrusive to allow the system booting reasonably.

The newly introduced PCD will disable the logging by default to allow system to continue using verbose without further negatively impacting the booting process.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change was tested on QEMU Q35 virtual platform.

## Integration Instructions

Platform should leave the newly introduced PCD as is to use the default disable setting. Should the platform want to enable syscall usage analysis, one can enable it by setting `gMmSupervisorPkgTokenSpaceGuid.PcdEnableSyscallLogs|TRUE` for analysis purposes.

### DIFF
--- a/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
+++ b/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
@@ -633,5 +633,6 @@ Exit:
   if (FeaturePcdGet (PcdEnableSyscallLogs)) {
     DEBUG ((DEBUG_INFO, "%a Exit...\n", __FUNCTION__));
   }
+
   return Ret;
 }
--- a/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
+++ b/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
@@ -238,7 +238,7 @@ SyscallDispatcher (
     mPcdCheck     = FALSE;
   }
 
-  if (FixedPcdGetBool (PcdEnableSyscallLogs)) {
+  if (FeaturePcdGet (PcdEnableSyscallLogs)) {
     while (!AcquireSpinLockOrFail (mCpuToken)) {
       CpuPause ();
     }
@@ -630,7 +630,7 @@ Exit:
     CpuDeadLoop ();
   }
 
-  if (FixedPcdGetBool (PcdEnableSyscallLogs)) {
+  if (FeaturePcdGet (PcdEnableSyscallLogs)) {
     DEBUG ((DEBUG_VERBOSE, "%a Exit...\n", __FUNCTION__));
   }
   return Ret;
--- a/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
+++ b/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
@@ -244,7 +244,7 @@ SyscallDispatcher (
     }
 
     DEBUG ((
-      DEBUG_VERBOSE,
+      DEBUG_INFO,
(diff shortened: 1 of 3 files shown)


MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
```diff
@@ -238,23 +238,25 @@ SyscallDispatcher (
...
-    __FUNCTION__,
-    CallIndex,
-    Arg1,
-    Arg2,
-    Arg3,
-    CallerAddr,
-    Ring3StackPointer
-    ));
+    DEBUG ((
+      DEBUG_VERBOSE,
```
makubacki, line 247: Given the message is opted into now with the PCD, do you think this should be `DEBUG_INFO`?

That would print the message if either info or verbose level is enabled + PCD is true as opposed to just verbose + PCD is true.

makubacki (approved):

MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
kuqin12: Yeah, I was thinking the same thing :) Will update shortly.

kenlautner (approved):

PR #152, opened 2023-07-10
MmSupervisorPkg: DumpInfo to more then stdout

Javagedes:
## Description

Updates DumpInfo function for a policy to dump it's contents to any IO object, not just to stdout. By default, the DumpInfo function will dump to stdout, keeping existing logic, but allows the caller to specify any IO object via the `outfs` optional parameter.

Additionally changes the dump for the purpose of the SupervisorPolicyMaker helper script to dump to logging.DEBUG rather than as a print statement, as the print statement does not end up in the log file.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Virtual Platforms

## Integration Instructions

N/A

### DIFF
--- a/MmSupervisorPkg/SupervisorPolicyTools/SupervisorPolicyMaker.py
+++ b/MmSupervisorPkg/SupervisorPolicyTools/SupervisorPolicyMaker.py
@@ -10,6 +10,7 @@
 ##
 
 import logging
+import io
 from argparse import ArgumentParser
 import os
 import datetime
@@ -164,13 +165,15 @@ class SupervisorPolicyMaker(IUefiHelperPlugin):
             ParseXmlAndAddToPolicy(xml_file_path, Policy)
 
             # print out our policy
-        print("=================================================")
-        print("========    Start Dumping Policy    =============")
-        print("=================================================")
-        Policy.DumpInfo(prefix="  ")
-        print("=================================================")
-        print("========    End Dumping Policy    ===============")
-        print("=================================================")
+        logging.debug("=================================================")
+        logging.debug("========    Start Dumping Policy    =============")
+        logging.debug("=================================================")
+        outfs = io.StringIO()
+        Policy.DumpInfo(prefix="  ", outfs = outfs)
+        logging.debug(outfs.getvalue())
+        logging.debug("=================================================")
+        logging.debug("========    End Dumping Policy    ===============")
+        logging.debug("=================================================")
         # write out policy as binary file
         if output_binary_path is not None:
             with open(output_binary_path, "wb") as f:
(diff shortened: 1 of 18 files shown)


MmSupervisorPkg/SupervisorPolicyTools/SupervisorPolicyMaker.py
```diff
@@ -164,13 +165,15 @@ def MakeSupervisorPolicy(output_version=Supervisor_Policy.FLEXBILE_STRUCTURE_VER
...
-        print("========    Start Dumping Policy    =============")
-        print("=================================================")
-        Policy.DumpInfo(prefix="  ")
-        print("=================================================")
-        print("========    End Dumping Policy    ===============")
-        print("=================================================")
+        logging.debug("=================================================")
+        logging.debug("========    Start Dumping Policy    =============")
+        logging.debug("=================================================")
+        outfs = io.StringIO()
```
kuqin12, line 171: Hmm, I would just keep it simple that the dumpinfo will also use logging.** functions.

MmSupervisorPkg/SupervisorPolicyTools/policy_entry.py
```diff
@@ -166,14 +167,14 @@ def Decode(self, Buffer) -> bytes:
...
+    def DumpInfo(self, prefix: str = "", outfs: IO = sys.stdout) -> None:
         '''Print the object in human readable form.
 
         prefix: str printed before each line of output.  Best used for whitespace and alignment
         '''
-        print(f"{prefix}Version: 0x{self.Version:08X}")
-        print(f"{prefix}Type: {self.Type}")
-        print(f"{prefix}EntrySize: {self.EntrySize}")
+        outfs.write(f"{prefix}Version: 0x{self.Version:08X}\n")
+        outfs.write(f"{prefix}Type: {self.Type}\n")
```
kuqin12, line 176: As mentioned, l think a logging.verbose should do this one, too.

MmSupervisorPkg/SupervisorPolicyTools/SupervisorPolicyMaker.py
Javagedes: There are two reasons I went this route:

1. This method allows a caller to change a single line of code (`outfs=<something`) to dump to anything - stdout, stderror, a file, a log, etc). Changing it to hard code `logging.*` reduces it's reusability. Someone would have to completely rewrite / change 50+ lines of code to write to a different logging level or to a file
2. This method allows the `Print()` function to continue working as expected. If we hardcode it the way you are suggesting, we break the Print() function that exists for each of these objects. It would now just log it instead of print it.

If you are OK with both of these situations, I can go through and change it to logging.DEBUG or logging.VERBOSE

MmSupervisorPkg/SupervisorPolicyTools/policy_entry.py
Javagedes: verbose is not a logging level as we only use the python provided logging levels. logging.DEBUG is the most detailed of the levels. Are there any changes you would like?

kuqin12 (approved):

makubacki (approved):

PR #156, opened 2023-07-21
Adding extra step of ownership inspcetion for buffer validity check

kuqin12:
## Description

Current implementation of `MmIsBufferOutsideMmValid` for the user instance will only check to see if the requested space is unblocked from the perspective of MM supervisor, which could result in the issue that the region is only unblocked at the supervisor level is still not accessible from user space. If this is the case, `FALSE` should be returned.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change is tested on Q35 platform and booted to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
+++ b/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
@@ -602,6 +602,14 @@ SyscallDispatcher (
       break;
     case SMM_MM_UNBLOCKED:
       Ret = (UINT64)MmIsBufferOutsideMmValid ((EFI_PHYSICAL_ADDRESS)Arg1, Arg2);
+      if (Ret != FALSE) {
+        // Add another check to see if the buffer is in the user space.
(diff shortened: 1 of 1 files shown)


MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
makubacki, line 1: I suggest the following PR description as it is a little easier to understand the problem to me:

Current implementation of `MmIsBufferOutsideMmValid` for the user instance will only check to see if the requested space is unblocked from the perspective of MM supervisor, which could result in the issue that the region is only unblocked at the supervisor level is still not accessible from user space. If this is the case, `FALSE` should be returned.

makubacki (approved):

MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c
kuqin12: Thanks, updated per suggestion.

kenlautner (approved):

PR #241, opened 2024-03-04
SpamPkg: Add GenAux tool

Javagedes:
## Description

Adds a rust crate that will generate the auxiliary files for the spam. Replaces the old python script.

**NOTE**: For this tool to execute properly, the gen_aux crate must be added to the repository's Cargo.toml's members list.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified CI builds

## Integration Instructions

The rust crate must be added to the workspace's cargo.toml file.


SpamPkg/Tools/gen_aux/Cargo.toml
kuqin12, line 1: Super dumb question, will this tool be built during pre-build time? or we need to build it prior to normal usage?

SpamPkg/Tools/gen_aux/src/auxgen.rs
```diff
@@ -0,0 +1,343 @@
+//! A module that contains the C-equivalent structs and final functions to
```
kuqin12, line 1: Probably want to call out what/where they are?

kuqin12 (approved):

SpamPkg/Tools/gen_aux/Cargo.toml
Javagedes: `cargo run` will compile and execute the binary. Since this ultimately will *not* be a build rule and build tool, and will only run on the release pipeline, I think that suffices. When I write the pipeline, I do plan on caching the build output of the tool in the pipeline, so that future builds will be very quick. We will see if that goes well when I get to that point

SpamPkg/Tools/gen_aux/src/auxgen.rs
Javagedes: Added !

PR #255, opened 2024-03-12
SpamBinBuild: Patch aux file hash into Stm.bin

Javagedes:
## Description

Adds the framework to consume and patch a pcd that represents the 32 byte hash of the auxiliary file. This includes a new patchable pcd, PcdAuxBinHash, and the post-build logic to patch the hash into the stm binary.

**NOTE**: The platform build will fail in the newly added step as the pcd gets optimized out of the binary due to not being used.

You can force it to be used by adding the following code to `StmInit.c`

L26: 

``` c
volatile UINT8                  *mMyValue;
```

L805:

``` c 
  mMyValue = PatchPcdGet8(PcdAuxBinHash);
  if (mMyValue != 0) {
    DEBUG((EFI_D_INFO, "PcdAuxBinHash - %x\n", (UINTN)mMyValue));
  }
```

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


SpamPkg/SpamBinBuild.py
```diff
@@ -249,15 +246,129 @@ def SetPlatformEnv(self):
...
+    if not map.exists():
+        raise FileNotFoundError(map)
+    if not dll.exists():
+        raise FileNotFoundError(dll)
+
+    cmd = "GenPatchPcdTable"
+    args = f"-m {map} -e {dll}"
+    RunCmd(cmd, args)
+
+    pcd_table = map.with_suffix(".BinaryPcdTable.txt")
```
kuqin12, line 334: Should we have a txt file that has a list of supported aux files somewhere?

Javagedes: What do you mean supported aux files? The pcd_table variable you put this comment on is just an artifact that the edk2 maintained tool `GenPatchPcdTable`, so that name is hardcoded.

If you mean the actual aux file, It should be such that any aux file generated by `aux_gen` should be supported on the firmware side.

kuqin12 (approved):

SpamPkg/Tools/GenSpamArtifacts/GenSpamArtifacts.py
```diff
@@ -0,0 +1,203 @@
...
+    Returns:
+        The hash of the auxiliary file.
+
+    Raises:    
+        FileNotFoundError: The file does not exist
+    """
+    if not file.exists():
+        raise FileNotFoundError(file)
+
+    hasher = hashlib.new("sha256")
```
kuqin12, line 112: nit: Given this is hardcoded to sha256, we probably should note this more explicitly somewhere.

Javagedes: Updated to mention the algorithm used is SHA256 in SpamPkg.dec, Stm.inf, and in GenSpamArtifacts.py

PR #277, opened 2024-04-11
Move secure policy into a common library

kuqin12:
# Preface

Please ensure you have read the [contribution docs](https://github.com/microsoft/mu/blob/master/CONTRIBUTING.md) prior
to submitting the pull request. In particular,
[pull request guidelines](https://github.com/microsoft/mu/blob/master/CONTRIBUTING.md#pull-request-best-practices).

## Description

The preparation and walk through of general policy as well as memory policy are common code that does not require module specific knowledge.

This change moves the policy related manipulation code into a common library implementation.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU Q35 platform and booted to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/MmSupervisorPkg/Core/MmSupervisorCore.c
+++ b/MmSupervisorPkg/Core/MmSupervisorCore.c
@@ -14,13 +14,13 @@
 #include "Handler/Handler.h"
 #include "PrivilegeMgmt/PrivilegeMgmt.h"
 #include "Telemetry/Telemetry.h"
-#include "Policy/Policy.h"
 #include "Test/Test.h"
 
 #include <Protocol/MmBase.h>
 #include <Protocol/PiPcd.h>
 
 #include <Guid/MmCommonRegion.h>
+#include <Library/SecurePolicyLib.h>
 
 EFI_STATUS
 MmCoreFfsFindMmDriver (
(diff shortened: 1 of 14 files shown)


MmSupervisorPkg/Library/SecurePolicyLib/MemPolicy.c
```diff
@@ -49,6 +51,22 @@ typedef union {
...
+
+  @param Buffer  The buffer start address to be checked.
+  @param Length  The buffer length to be checked.
+
+  @retval TRUE  This buffer is not part of MMRAM.
+  @retval FALSE This buffer is from MMRAM.
+**/
+BOOLEAN
+EFIAPI
+IsBufferInsideMmram (
```
kuqin12, line 65: This is a catch: the module that uses this needs to have its own implementation. As per scope of this initiative, the supv core has an implementation based on global variables, whereas the spam core will always query the MSR to get the value. This is because the same routine will run on all cores, so saving to global for spam core is not as safe. We run its validation routine once anyway.

Thoughts?

Javagedes (approved):

MmSupervisorPkg/Core/MmSupervisorCore.c
```diff
@@ -798,6 +799,133 @@ DiscoverStandaloneMmDriversInFvHobs (
...
+        DEBUG_ERROR,
+        "[%a] Failed to find raw section from discovered policy file - %r\n",
+        __FUNCTION__,
+        Status
+        ));
+      break;
+    }
+
+    PolicySize = ((SMM_SUPV_SECURE_POLICY_DATA_V1_0 *)SectionData)->Size;
+    if (PolicySize > SectionDataSize) {
```
kenlautner, line 870: Why do we only care if the Policy Size is larger than the SectionDataSize?  I just want more information on what we're checking here.

kuqin12: This means that the policy binary we carried in the FV is smaller than the policy data indicates. This normally means that the policy data is incorrect, but could also mean that the system is under attack.

kenlautner: I understand why that might be a problem.  I was more wondering about why it's okay not to check if the PolicySize is smaller than SectionDataSize.  Wouldn't that also mean there is an error or there was tampering?

kuqin12: It might mean that there some bytes that we do not understand. But if that is the case, the security check below will call out what bytes are unexpected.

kenlautner (approved):

MmSupervisorPkg/Library/SecurePolicyLib/MemPolicy.c
kuqin12: Seems to be fine for now.

PR #315, opened 2024-07-13
Fixing override validation for basecore 2023110009.0.0

apop5:
## Description

Basecore 2023110009.0.0 included the change for introducing panic lib. 

MM Supervisor is not switching to consume panic lib.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
N/A

## Integration Instructions
N/A


MmSupervisorPkg/Core/MmSupervisorCore.inf
makubacki, line 1: The statement in the PR description that "MM Supervisor is not switching to consume panic lib." is confusing. It is already using PanicLib - https://github.com/microsoft/mu_feature_mm_supv/blob/a87979f0bfb5637ac8179573fc93d6fc65bf5c73/MmSupervisorPkg/Core/Relocate/Relocate.c#L602.

I think there's a few other places that also make sense for it to be used in MmSupervisorCore, but that's not important for pre-existing code in the context of this override hash update.

apop5: Based on offline conversations, updated to call PanicLib functions in place of previous CpuDeadloop calls.

kuqin12 (approved):

MmSupervisorPkg/Core/Relocate/Relocate.c
```diff
@@ -1433,7 +1433,7 @@ FindSmramInfo (
         ResetWarm ();
       }
 
-      CpuDeadLoop ();
+      PANIC ("Critical HOB missing that describes MMRAM regions\n");
```
makubacki, line 1436: This panic string adds a new line where the others do not. I don't think it's a big deal either way. Just wanted to call it out in case you would like it to be consistent. The `PANIC()` API states the implementation will add a newline.

makubacki (approved):

MmSupervisorPkg/Core/MmSupervisorCore.inf
makubacki: Thanks

MmSupervisorPkg/Core/Relocate/Relocate.c
apop5: Thank you for catching the inconsistency. It has been corrected in the latest push.

PR #386, opened 2025-01-28
Separate .bss section from the .data sections in the RELEASE image

kenlautner:
## Description

Separate the .bss section of the image in RELEASE builds.  This is because we typically merge all of the different .data sections in RELEASE builds which normally includes the .bss section.  This breaks SEA validation as it expects the .bss section to be ignored during validation but instead we're including it in the normal .data section validation.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on a SEA physical platform.  Validation completes successfully with the change.

## Integration Instructions

N/A


MmSupervisorPkg/Core/MmSupervisorCore.inf
```diff
@@ -234,7 +234,7 @@
 [BuildOptions.common]
   #Subsystem version will be used as MmSupervisor driver version
   #Image version will be used as MmSupervisor security patch level(SPL) value
-  MSFT:RELEASE_*_*_DLINK_FLAGS  = /DLL /SUBSYSTEM:CONSOLE,$(RELEASE_DRIVER_VERSION) /VERSION:$(MAJOR_PATCH_LEVEL).$(MINOR_PATCH_LEVEL)
+  MSFT:RELEASE_*_*_DLINK_FLAGS  = /DLL /SUBSYSTEM:CONSOLE,$(RELEASE_DRIVER_VERSION) /VERSION:$(MAJOR_PATCH_LEVEL).$(MINOR_PATCH_LEVEL) /MERGE:.bss=.none
```
makubacki, line 237: With this, do you see the .bss section called ".none" in the binary?

kenlautner: In the map file it still appears as .bss and class type DATA.

makubacki (approved):

kuqin12 (approved):

PR #390, opened 2025-02-03
MmSupervisorPkg: Add MmSupervisorRing3Performance

### LINKED ISSUE #383: [Feature]: Add MM Performance Record Support
### Feature Overview

Currently, the supervisor does not support a `PerformanceLib` instance that will return MM performance records.

### Solution Overview

Add support to the MM Supervisor for extracting MM performance records.

### Alternatives Considered

_No response_

### Urgency

Medium

Someone else needs to implement the feature

No maintainer feedback needed

### Anything else?

I may be able to implement this feature. I don't have bandwidth to work on it right away.

makubacki:
## Description

Closes #383 

Adds a new `MM_STANDALONE` driver that is used to link against a
`PerformanceLib` core instance for MM user mode drivers.

An empty driver is reserved for this use case for the following
reasons:

1. Make integration and debug obvious and easy by having a clearly
   named and designated driver for this purpose.
2. To be among the first MM user mode drivers dispatched to install
   the performance protocol and make the performance service
   available to other user mode drivers as early as possible.
3. To avoid linking directly to `MmSupervisorRing3Broker`. The MM
   Core performance library instance code has functionality that
   depends on MM services to be available. Since the Ring 3 Broker
   is reponsible for initializing MM services in user mode, it is
   complicated and error prone to attempt to defer using those
   services to install the performance protocol until syscall has
   been set up with the core and a proper `gMmst` global could be
   used by the common MM code. This is all much simpler and easier
   to integrate in a separate MM user mode driver.
4. To avoid linking the library to a pre-existing MM module that
   may have an unknown number of depdencies in its DEPEX. Since
   the DEPEX on a given platform aggregates all of the dependencies
   in linked libraries, it is not safe to assume any given MM module
   not dedicated exclusively for this purpose will not have unrelated
   dependencies that unnecessarily prolong installation of the
   performance protocol.

---

Temporarily includes this commit:

**TEMP - .pytool/CISettings.py: Use Mu Basecore Standalone MM Perf Branch**

Temporarily use this branch for testing.

This commit will be reverted after the following Mu Basecore PR is
merged:

https://github.com/microsoft/mu_basecore/pull/1268

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- Use `MmSupervisorRing3Performance` to retrieve MM performance records
  with the MM Supervisor on Q35.

## Integration Instructions

- Included in MmSupervisorPkg/Docs/PlatformIntegration/PlatformIntegrationSteps.md

### DIFF
--- a/MmSupervisorPkg/Docs/PlatformIntegration/PlatformIntegrationSteps.md
+++ b/MmSupervisorPkg/Docs/PlatformIntegration/PlatformIntegrationSteps.md
@@ -285,6 +285,9 @@ flash drivers, SW MMI dispatcher drivers, etc.
   IoLib|MmSupervisorPkg/Library/BaseIoLibIntrinsicSysCall/BaseIoLibIntrinsic.inf
   SysCallLib|MmSupervisorPkg/Library/SysCallLib/SysCallLib.inf
 
+  # This library instance is only necessary if performance tracing is enabled in MM code.
+  PerformanceLib|MdeModulePkg/Library/SmmPerformanceLib/StandaloneMmPerformanceLib.inf
+
 [Components.IA32]
   MmSupervisorPkg/Drivers/StandaloneMmHob/StandaloneMmHob.inf
   MmSupervisorPkg/Drivers/MmCommunicationBuffer/MmCommunicationBufferPei.inf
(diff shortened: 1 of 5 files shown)


.pytool/CISettings.py
```diff
@@ -189,7 +189,7 @@ def GetDependencies(self):
             {
                 "Path": "MU_BASECORE",
                 "Url": "https://github.com/microsoft/mu_basecore.git",
-                "Branch": "dev/202405"
+                "Branch": "feature/add_standalonemm_perf_lib"
```
apop5, line 192: Comment is just a reminder to restore the basecore.

MmSupervisorPkg/Docs/PlatformIntegration/PlatformIntegrationSteps.md
```diff
@@ -312,7 +315,19 @@ flash drivers, SW MMI dispatcher drivers, etc.
       # Note that this should be whatever suits the target platform + MM standalone conversion for constructor input arguments
       SmmCpuFeaturesLib|$(PLATFORM_SI_PACKAGE)/Library/SmmCpuFeaturesLib/SmmCpuFeaturesLib.inf
   }
-  MmSupervisorPkg/Drivers/MmSupervisorRing3Broker/MmSupervisorRing3Broker.inf
+  MmSupervisorPkg/Drivers/MmSupervisorRing3Broker/MmSupervisorRing3Broker.inf {
```
apop5, line 318: With Ring3 being the first to load, this means that ring 3 cannot support performance libs.

Would it be better to just remove the PerformanceLib libraryclasses and usage from ring 3? It would simplify the consumption because it will not require projects to override the PerformanceLib in their dsc files.

apop5 (approved):

MmSupervisorPkg/Docs/PlatformIntegration/PlatformIntegrationSteps.md
makubacki: There was an intention to keep the integration process as familiar as possible and share implementation between supervised and non-supervised Standalone MM. Today, `PerformanceLib` has a core instance and a non-core instance in a DXE or MM environment.

Supervised MM presents the following complications:

- `MmSupervisorCore` (needs separate perf for ring 0 data)
- `MmSupervisorRing3Broker` (responsible for setting up ring 3 - would need special handling to cache perf records if desired since it creates a race condition to MM services being available used by other user drivers).
- `MmSupervisorRing3Performance` (first user MM module to link user perf init to - core lib)
- Everything else can use the normal `StandaloneMmPerformanceLib` (non-core instance)

Since the reasons perf init is being bumped out to `Ring3Performance` is not unique to perf, I thought it should not be too unexpected and it was relatively simple to just link the core instance against the first valid user driver (which would have to happen anyway in some form) and then use the non-core instance for other MM modules (which would also need to happen in some form).

Perf from `MmSupervisorCore` and `Ring3Broker` need additional work as noted but I think that will fit into this fairly naturally when it is done.

kuqin12 (approved):

PR #394, opened 2025-02-10
gen_aux: Add support for requiring all symbols have a rule

Javagedes:
## Description

Adds a configuration file config option `no_missing_rules` that when set, will cause the tool to exit with an error if any symbol does not have a rule in the configuration file.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated the "status quo" did not change, but that setting the flag to true results in a build break that prints the symbol and its address.

## Integration Instructions

Consumers may set the `no_missing_rules` to true to enable this functionality.


kenlautner (approved):

makubacki:
Doesn't this result in a breaking change?

>> Consumers must add a `[config]` section header above any currently existing `autogen` configuration
>>
>> e.g.
>>
>> ``` toml
>> autogen = false
>> ```
>>
>> to
>>
>> ``` toml
>> [config]
>> autogen = false
>> ```

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/auxgen.rs
```diff
@@ -241,10 +258,18 @@ impl AuxBuilder {
...
+            if file.config.auto_gen && file.config.no_missing_rules {
+                return Err(anyhow::anyhow!("auto_gen and no_missing_symbols cannot be true at the same time."));
+            }
+            
+            self.exit_on_missing_rules = file.config.no_missing_rules;
+            self.auto_generate = file.config.auto_gen;
+            self.excluded_symbols = file.config.excluded_symbols;
+            self.rules = file.rules;
+            self.key_symbols = file.key_symbols;
+
```
makubacki, line 272: Nit: Was this empty line intentional?

```suggestion
```

makubacki (approved):

Javagedes:
> Doesn't this result in a breaking change?
> 
> > > Consumers must add a `[config]` section header above any currently existing `autogen` configuration
> > > e.g.
> > > ```toml
> > > autogen = false
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > to
> > > ```toml
> > > [config]
> > > autogen = false
> > > ```

It is. I added Integration instructions but did not mark it as a breaking change. I've updated. Good catch.

PR #396, opened 2025-02-10
Make it so PiSmmCpuSmiEntryFixupAddress is only called when not using SEA

kenlautner:
## Description

Currently we touch up the SmiEntry.nasm file when calling PiSmmCpuSmiEntryFixupAddress which is something we don't want to do when using the SEA  MmiEntrySea.nasmb file.  This change makes it so PiSmmCpuSmiEntryFixupAddress is only called when we aren't using a different MmiEntry._>

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on physical system.  The SmiEntry.nasm is no longer being fixed up.

## Integration Instructions

N/A

### DIFF
--- a/MmSupervisorPkg/Core/Relocate/Relocate.c
+++ b/MmSupervisorPkg/Core/Relocate/Relocate.c
@@ -697,8 +697,12 @@ SetupSmiEntryExit (
 
   //
   // Initialize address fixup
-  //
-  PiSmmCpuSmiEntryFixupAddress ();
+  //s
+
+  // This is only necessary when using the MmSupervisorCore SmiEntry.nasm.  Ignore it with SEA.
+  if (SmmCpuFeaturesGetSmiHandlerSize () == 0) {
+    PiSmmCpuSmiEntryFixupAddress ();
+  }
 
   //
   // Initialize Debug Agent to support source level debug in SMM code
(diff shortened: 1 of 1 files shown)


MmSupervisorPkg/Core/Relocate/Relocate.c
```diff
@@ -698,7 +698,11 @@ SetupSmiEntryExit (
   //
   // Initialize address fixup
   //
-  PiSmmCpuSmiEntryFixupAddress ();
+
+  // This is only necessary when using the MmSupervisorCore SmiEntry.nasm.  Ignore it with SEA.
```
kuqin12, line 702: nit: you probably do not want to mention SEA in the comment. Just that if feature lib has its own entry code, no fixup is needed.

kuqin12 (approved):

makubacki (approved):

MmSupervisorPkg/Core/Relocate/Relocate.c
kenlautner: Okay.  I fixed the comment to be implementation agnostic.

PR #401, opened 2025-02-14
[REBASE&FF] image_validation: bugfix aux file bounds checks

Javagedes:
## Description

This commit fixes two separate issues with the bounds checks in the image validation code that ensures certain pieces of data in the auxiliary file are within the bounds of the file, according to the size expectations described in the validation entry headers.

1. The first issue resolved is specific to the IMAGE_VALIDATION_CONTENT validation type. In this validation type, the header contains N number of additional bytes, that are the expected value that the particular content should be. The purpose of this validation is to ensure that the content provided by the auxiliary file for validation does not extend beyond the bounds of the file.

The issue was that the bounds check only validated that the overall size of the header + content was less than the overall size of the auxiliary file. This has been updated to ensure that:

addr_of(header) + size_of(header) + size_of(content) <= addr_of(aux) + size_of(aux)

2. The second is that only one validation type had a bounds check to ensure the default value that we revert data to was within the bounds of the auxiliary file. This has been moved out of the specific validation type and is now a general check immediately before the reversion of the particular data.

- [ ] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

TestApp Continues to pass

## Integration Instructions

N/A

### DIFF
--- a/SeaPkg/Core/Runtime/SeaResponderReport.c
+++ b/SeaPkg/Core/Runtime/SeaResponderReport.c
@@ -104,10 +104,10 @@ Done:
 EFI_STATUS
 EFIAPI
 VerifyAndHashImage (
-  IN  UINTN                         ImageBase,
+  IN  EFI_PHYSICAL_ADDRESS          ImageBase,
   IN  UINT64                        ImageSize,
   IN  IMAGE_VALIDATION_DATA_HEADER  *AuxFileHdr,
-  IN  UINT64                        PageTableBase,
+  IN  EFI_PHYSICAL_ADDRESS          PageTableBase,
   OUT TPML_DIGEST_VALUES            *DigestList
   )
 {
@@ -321,10 +321,10 @@ SeaResponderReport (
   UINT8                             *LocalMmiEntryBase = NULL;
   PER_CORE_MMI_ENTRY_STRUCT_HDR     *MmiEntryStructHdr;
   UINT32                            MmiEntryStructHdrSize;
-  UINT64                            MmSupervisorBase;
+  EFI_PHYSICAL_ADDRESS              MmSupervisorBase;
   UINT64                            MmSupervisorImageSize;
   UINT64                            FirmwarePolicyBase;
-  UINT64                            SupvPageTableBase;
+  EFI_PHYSICAL_ADDRESS              SupvPageTableBase;
   TPML_DIGEST_VALUES                DigestList;
   UINT8                             *DrtmSmmPolicyData;
   SMM_SUPV_SECURE_POLICY_DATA_V1_0  *FirmwarePolicy;
@@ -508,7 +508,7 @@ SeaResponderReport (
   }
 
   // CR3 should be pointing to the page table from symbol list in the aux file
-  SupvPageTableBase = *(UINT64 *)(MmSupervisorBase + PageTableSymbol->Offset);
+  SupvPageTableBase = *(EFI_PHYSICAL_ADDRESS *)(MmSupervisorBase + PageTableSymbol->Offset);
   if (Fixup32Ptr[FIXUP32_CR3_OFFSET] != SupvPageTableBase) {
     DEBUG ((DEBUG_ERROR, "%a Calculated page table 0x%p does not match MM entry code populated value 0x%p.\n", __func__, SupvPageTableBase, Fixup32Ptr[FIXUP32_CR3_OFFSET]));
     Status = EFI_SECURITY_VIOLATION;
--- a/SeaPkg/Core/Runtime/SeaResponderReport.c
+++ b/SeaPkg/Core/Runtime/SeaResponderReport.c
@@ -93,8 +93,7 @@ Done:
 
   @param[in] ImageBase      The base address of the image.
   @param[in] ImageSize      The size of the image.
-  @param[in] AuxFileBase    The base address of the auxiliary file.
-  @param[in] AuxFileLength  The length of the auxiliary file.
+  @param[in] AuxFileHdr     The header of the auxiliary file.
   @param[in] PageTableBase  The base address of the page table.
   @param[out] DigestList    The digest list of the image.
 
@@ -105,12 +104,11 @@ Done:
 EFI_STATUS
 EFIAPI
 VerifyAndHashImage (
-  IN  UINTN               ImageBase,
-  IN  UINT64              ImageSize,
-  IN  UINT64              AuxFileBase,
-  IN  UINT64              AuxFileLength,
-  IN  UINT64              PageTableBase,
-  OUT TPML_DIGEST_VALUES  *DigestList
+  IN  UINTN                         ImageBase,
+  IN  UINT64                        ImageSize,
+  IN  IMAGE_VALIDATION_DATA_HEADER  *AuxFileHdr,
+  IN  UINT64                        PageTableBase,
+  OUT TPML_DIGEST_VALUES            *DigestList
   )
 {
   EFI_STATUS                    Status;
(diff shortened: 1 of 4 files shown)


SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
```diff
@@ -1009,10 +997,25 @@ PeCoffImageDiffValidation (
         break;
     }
 
+    // Exit all validation checks if this validation entry fails
     if (EFI_ERROR (Status)) {
       break;
     }
 
+    // Exit all validation checks if this validation entry's default value overflows the Auxiliary file buffer.
+    if (ImageValidationEntryHdr->OffsetToDefault + ImageValidationEntryHdr->Size > ReferenceDataSize) {
```
kuqin12, line 1006: I think you want this check to occur earlier. We would not want to touch this region if this aux_bin is bad.

Javagedes: Sure that makes sense. I'll move this check to before we run the symbol validation code

Javagedes: moved to the top!

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
```diff
@@ -844,7 +857,8 @@ PeCoffImageDiffValidation (
         break;
       case IMAGE_VALIDATION_ENTRY_TYPE_CONTENT:
         ImageValidationEntryContent = (IMAGE_VALIDATION_CONTENT *)(ImageValidationEntryHdr);
-        if (sizeof (*ImageValidationEntryContent) + ImageValidationEntryHdr->Size > ReferenceDataSize) {
+        // Ensure "Content" in the header (TargetContent) does not overflow the Auxiliary file buffer.
+        if ((UINT8 *)ImageValidationEntryContent + sizeof (*ImageValidationEntryContent) + ImageValidationEntryHdr->Size > (UINT8 *)ReferenceData + ReferenceDataSize) {
```
makubacki, line 861: Not directly related to this PR, but is there a reason why the `ReferenceData` type is `VOID *` instead of `IMAGE_VALIDATION_DATA_HEADER *`? That would make the interface clearer in terms of what is expected.

makubacki: @Javagedes, the comment is helpful as-is, but can you please expand that the entry list is contained with the aux file? I think this would be more self-documented/obvious if `ReferenceData` would just `AuxFile` or `AuxData` per my understanding.

kuqin12: > Not directly related to this PR, but is there a reason why the `ReferenceData` type is `VOID *` instead of `IMAGE_VALIDATION_DATA_HEADER *`? That would make the interface clearer in terms of what is expected.

I think we could. I changed this interface quite a few times back when I wrote this. So I probably just dwelled on a version that worked :)

Javagedes: I will update the ReferenceData and ReferenceDataSize accordingly, along with looking into changing it from `VOID *` to `IMAGE_VALIDATION_DATA_HEADER *`

👍

Javagedes: @kuqin12  Should we be using the AuxData Size coming from what the `PcdGetSize` says, or should we be using the header field `Size` that is also the size of the entire aux data?

kuqin12: This information was intentionally passed in, so that this code can live in SeaPkg but linked to supervisor. Otherwise, supervisor will have dependency on these PCDs defined in SeaPkg.

Javagedes: I meant the start of the aux file, `IMAGE_VALIDATION_DATA_HEADER` has a size field that is the size of the aux file. I'm curious if we can verify that size against the pcd size as some point, and then just pass down _only_ the AuxData object, through the callstack, instead of auxData + size(where the size is coming from the `PcdGetSize `)

kuqin12: We can, but we will need to do it from the caller site, which will become take the PCD, verify size sanity, pass to this function.

makubacki: Thanks for looking into this too, it was another point that I also noticed could reduce some complexity in reviewing the function if addressed somewhere else.

SeaPkg/Core/Runtime/SeaResponderReport.c
```diff
@@ -105,12 +104,11 @@ Range1InsideRange2 (
 EFI_STATUS
 EFIAPI
 VerifyAndHashImage (
-  IN  UINTN               ImageBase,
-  IN  UINT64              ImageSize,
-  IN  UINT64              AuxFileBase,
-  IN  UINT64              AuxFileLength,
-  IN  UINT64              PageTableBase,
-  OUT TPML_DIGEST_VALUES  *DigestList
+  IN  UINTN                         ImageBase,
```
makubacki, line 107: Something else that sticks out about the function signature is that `ImageBase` is `UINTN` and `PageTableBase` is `UINT64` though they are both base addresses. Is there a reason both base addresses don't use the same type? It would be clearer to use `EFI_PHYSICAL_ADDRESS` which is a `UINT64`.

Javagedes: Sure, I can update this to `EFI_PHYSICAL_ADDRESS` so long as @kuqin12 does not have a specific reason these are different

Javagedes: Updated in the last commit. Can drop if need be, depending on kuqin12.

SeaPkg/Core/Runtime/SeaResponderReport.c
```diff
@@ -367,27 +366,34 @@ SeaResponderReport (
...
+  if (AuxFileHdr == NULL) {
     DEBUG ((DEBUG_ERROR, "%a Reported aux file base address is NULL!\n", __func__));
     Status = EFI_SECURITY_VIOLATION;
     goto Exit;
   }
 
-  if (!IsBufferInsideMmram (AuxFileBase, AuxFileSize)) {
-    DEBUG ((DEBUG_ERROR, "%a Reported aux file (0x%p: 0x%x) does not reside in MMRAM region!\n", __func__, AuxFileBase, AuxFileSize));
+  if (AuxFileHdr->Size > AuxFileSize) {
+    DEBUG ((DEBUG_ERROR, "%a Reported aux file size is larger than the actual size %x %x\n", __func__, AuxFileHdr->Size, AuxFileSize));
```
makubacki, line 377: nit: `0x` preceding the hexadecimal value can reduce confusion in interpreting results in logs.

```suggestion
    DEBUG ((DEBUG_ERROR, "%a Reported aux file size is larger than the actual size 0x%x 0x%x\n", __func__, AuxFileHdr->Size, AuxFileSize));
```

makubacki (approved):

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
Javagedes: I just confirmed the TestApp did not fail. I will close this.

SeaPkg/Core/Runtime/SeaResponderReport.c
kuqin12: I think updating to efi_physical_address makes sense! Thanks for fixing it!

kuqin12 (approved):

PR #404, opened 2025-02-19
SeaPkg: Make Image Validation Testable

Javagedes:
## Description

Breaks away the image validation into a standalone library, exposing the validation functions to be easily testable.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Passed the TestApp

## Integration Instructions

Platforms will need to add `PeCoffValidationLib|SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.inf` to their DSC

### DIFF
--- a/SeaPkg/Include/Library/PeCoffValidationLib.h
+++ b/SeaPkg/Include/Library/PeCoffValidationLib.h
@@ -4,7 +4,7 @@
   multiple validation entries. A validation entry specifies a specific region in the
   target image buffer to validate against, a validation type to perform, and any
   necessary data to assist in the validation.
-  
+
   There currently exists five validation types:
   - Non-Zero: Validates that the specified region in the target image buffer is not all zero.
   - Content: Validates that the specified region in the target image buffer matches the content in the reference data.
@@ -18,6 +18,7 @@
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
+
 #ifndef BASE_PECOFF_VALIDATION_LIB_H_
 #define BASE_PECOFF_VALIDATION_LIB_H_
 
@@ -33,9 +34,9 @@
 **/
 EFI_STATUS
 PeCoffImageValidationNonZero (
-    IN VOID                           *TargetImage,
-    IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr
-);
+  IN VOID                           *TargetImage,
+  IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr
+  );
 
 /**
   Validates a specific region in the target image buffer denoted by [Hdr->Offset: Hdr->Offset + Hdr->Size]
@@ -51,10 +52,10 @@ PeCoffImageValidationNonZero (
 **/
 EFI_STATUS
 PeCoffImageValidationContent (
-  IN VOID                               *TargetImage,
-  IN IMAGE_VALIDATION_ENTRY_HEADER      *Hdr,
-  IN CONST IMAGE_VALIDATION_DATA_HEADER *ImageValidationHdr
-);
+  IN VOID                                *TargetImage,
+  IN IMAGE_VALIDATION_ENTRY_HEADER       *Hdr,
+  IN CONST IMAGE_VALIDATION_DATA_HEADER  *ImageValidationHdr
+  );
 
 /**
   Validates a specific region in the target image buffer denoted by [Hdr->Offset: Hdr->Offset + Hdr->Size]
@@ -64,7 +65,7 @@ PeCoffImageValidationContent (
   @param[in] TargetImage    The pointer to the target image buffer.
   @param[in] Hdr            The header of the validation entry.
   @param[in] PageTableBase  The base address of the page table.
-    
+
   @returns EFI_SUCCESS             The target image passes the validation.
   @returns EFI_INVALID_PARAMETER   The validation entry has invalid must have and must not have attributes.
   @returns EFI_INVALID_PARAMETER   The validation entry data size is invalid. It must be a pointer size.
@@ -75,7 +76,7 @@ PeCoffImageValidationMemAttr (
   IN VOID                           *TargetImage,
   IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr,
   IN EFI_PHYSICAL_ADDRESS           PageTableBase
-);
+  );
 
 /**
   Validates a specific region in the target image buffer denoted by [Hdr->Offset: Hdr->Offset + Hdr->Size]
@@ -94,7 +95,7 @@ PeCoffImageValidationSelfRef (
   IN VOID                           *TargetImage,
   IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr,
   IN VOID                           *OriginalImageBaseAddress
-);
+  );
 
 /**
   Validates a specific region in the target image buffer denoted by [Hdr->Offset: Hdr->Offset + Hdr->Size]
@@ -111,6 +112,6 @@ EFI_STATUS
 PeCoffImageValidationPointer (
   IN VOID                           *TargetImage,
   IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr
-);
+  );
 
 #endif // BASE_PECOFF_VALIDATION_LIB_H_
--- a/SeaPkg/Include/Library/PeCoffValidationLib.h
+++ b/SeaPkg/Include/Library/PeCoffValidationLib.h
@@ -63,7 +63,7 @@ PeCoffImageValidationContent (
 
 /**
   Validates a specific region in the target image buffer denoted by [Hdr->Offset: Hdr->Offset + Hdr->Size]
-  belongs to a user page that is mapped inside MM and teh page attributes match the requirements specified
+  belongs to a user page that is mapped inside MM and the page attributes match the requirements specified
   by the validation entry.
 
   @param[in] TargetImage    The pointer to the target image buffer.
--- a/SeaPkg/Include/Library/PeCoffValidationLib.h
+++ b/SeaPkg/Include/Library/PeCoffValidationLib.h
@@ -35,8 +35,8 @@
 EFI_STATUS
 EFIAPI
 PeCoffImageValidationNonZero (
-  IN VOID                           *TargetImage,
-  IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr
+  IN CONST VOID                           *TargetImage,
+  IN CONST IMAGE_VALIDATION_ENTRY_HEADER  *Hdr
   );
 
 /**
@@ -54,9 +54,9 @@ PeCoffImageValidationNonZero (
 EFI_STATUS
 EFIAPI
 PeCoffImageValidationContent (
-  IN VOID                                *TargetImage,
-  IN IMAGE_VALIDATION_ENTRY_HEADER       *Hdr,
-  IN CONST IMAGE_VALIDATION_DATA_HEADER  *ImageValidationHdr
+  IN CONST VOID                           *TargetImage,
+  IN CONST IMAGE_VALIDATION_ENTRY_HEADER  *Hdr,
+  IN CONST IMAGE_VALIDATION_DATA_HEADER   *ImageValidationHdr
   );
 
 /**
@@ -76,9 +76,9 @@ PeCoffImageValidationContent (
 EFI_STATUS
 EFIAPI
 PeCoffImageValidationMemAttr (
-  IN VOID                           *TargetImage,
-  IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr,
-  IN EFI_PHYSICAL_ADDRESS           PageTableBase
+  IN CONST VOID                           *TargetImage,
+  IN CONST IMAGE_VALIDATION_ENTRY_HEADER  *Hdr,
+  IN EFI_PHYSICAL_ADDRESS                 PageTableBase
   );
 
 /**
@@ -96,9 +96,9 @@ PeCoffImageValidationMemAttr (
 EFI_STATUS
 EFIAPI
 PeCoffImageValidationSelfRef (
-  IN VOID                           *TargetImage,
-  IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr,
-  IN VOID                           *OriginalImageBaseAddress
+  IN CONST VOID                           *TargetImage,
+  IN CONST IMAGE_VALIDATION_ENTRY_HEADER  *Hdr,
+  IN CONST VOID                           *OriginalImageBaseAddress
   );
 
 /**
@@ -115,8 +115,8 @@ PeCoffImageValidationSelfRef (
 EFI_STATUS
 EFIAPI
 PeCoffImageValidationPointer (
-  IN VOID                           *TargetImage,
-  IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr
+  IN CONST VOID                           *TargetImage,
+  IN CONST IMAGE_VALIDATION_ENTRY_HEADER  *Hdr
   );
 
 #endif // PECOFF_VALIDATION_LIB_H_
--- a/SeaPkg/Include/Library/PeCoffValidationLib.h
+++ b/SeaPkg/Include/Library/PeCoffValidationLib.h
@@ -48,6 +48,8 @@ PeCoffImageValidationNonZero (
   @param[in] ImageValidationHdr  The pointer to the auxiliary file data buffer to assist.
 
   @retval EFI_SUCCESS             The target image passes the validation.
+  @retval EFI_INVALID_PARAMETER   One of the input parameters is a null pointer.
+  @retval EFI_INVALID_PARAMETER   The provided header has an invalid signature
   @retval EFI_COMPROMISED_DATA    The content to match against overflows the auxiliary file.
   @retval EFI_SECURITY_VIOLATION  The specified buffer in the target image does not match the reference data.
 **/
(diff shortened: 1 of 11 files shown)


SeaPkg/Include/Library/PeCoffValidationLib.h
```diff
@@ -0,0 +1,117 @@
...
+             specified by TargetOffset in the validation entry.
+  - Pointer: Validates that the specified region in the target image buffer is a pointer, and that the pointer is not NULL.
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef BASE_PECOFF_VALIDATION_LIB_H_
+#define BASE_PECOFF_VALIDATION_LIB_H_
```
makubacki, line 23:
```suggestion
#ifndef PECOFF_VALIDATION_LIB_H_
#define PECOFF_VALIDATION_LIB_H_

```

SeaPkg/Include/Library/PeCoffValidationLib.h
```diff
@@ -0,0 +1,117 @@
...
+
+  @param[in] TargetImage  The pointer to the target image buffer.
+  @param[in] Hdr          The header of the validation entry.
+
+  @return EFI_SUCCESS             The target image passes the validation.
+  @return EFI_SECURITY_VIOLATION  The specified buffer in the target image is all zero.
+**/
+EFI_STATUS
+PeCoffImageValidationNonZero (
+  IN VOID                           *TargetImage,
```
makubacki, line 37: I know these functions already existed but can some of the params be tightened up in interfaces where possible? Like `CONST VOID *` to better reflect usage.

SeaPkg/Include/Library/PeCoffValidationLib.h
```diff
@@ -0,0 +1,117 @@
...
+  @return EFI_INVALID_PARAMETER   The validation entry has invalid size.
+  @return EFI_SECURITY_VIOLATION  The target image does not match the content in the original image buffer.
+**/
+EFI_STATUS
+PeCoffImageValidationPointer (
+  IN VOID                           *TargetImage,
+  IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr
+  );
+
+#endif // BASE_PECOFF_VALIDATION_LIB_H_
```
makubacki, line 117:
```suggestion
#endif // PECOFF_VALIDATION_LIB_H_
```

SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.inf
```diff
@@ -0,0 +1,30 @@
...
+  MODULE_TYPE                    = BASE
+  VERSION_STRING                 = 1.0
+  LIBRARY_CLASS                  = PeCoffValidationLib
+
+#
+#  VALID_ARCHITECTURES           = IA32 X64 EBC ARM AARCH64
+#
+
+[Sources]
+  Validation.c
```
makubacki, line 22: Not a big deal, by normally this would match the inf file name. Is there a reason not to do so here?

SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.inf
```diff
@@ -0,0 +1,30 @@
+## @file
+#  PE/COFF Image Validation Library implementation.
```
makubacki, line 2: Because the library is in `SeaPkg` I think the generic (and more broadly named) `BasePeCoffValidationLib` would be understood to validate PE/COFF images within the scope of SEA. Here though, it might be good to reinforce that.

```suggestion
#  SEA PE/COFF Image Validation Library implementation.
```

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
```diff
@@ -0,0 +1,325 @@
...
+  IN  EFI_PHYSICAL_ADDRESS  PageTableBase,
+  IN  EFI_PHYSICAL_ADDRESS  BaseAddress,
+  IN  UINT64                Length,
+  OUT UINT64                *Attributes
+  );
+
+/**
+  Helper function that will evaluate the page where the input address is located belongs to a
+  user page that is mapped inside MM.
+  @param  Address           Target address to be inspected.
```
makubacki, line 41:
```suggestion
  Helper function that will evaluate the page where the input address is located belongs to a
  user page that is mapped inside MM.

  @param  Address           Target address to be inspected.
```

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
```diff
@@ -0,0 +1,325 @@
...
+  @param  Size              Address range to be inspected.
+  @param  IsUserRange       Pointer to hold inspection result, TRUE if the region is in User pages, FALSE if
+                            the page is in supervisor pages. Should not be used if return value is not EFI_SUCCESS.
+  @return     The result of inspection operation.
+**/
+EFI_STATUS
+InspectTargetRangeAttribute (
+  IN  EFI_PHYSICAL_ADDRESS  PageTableBase,
```
makubacki, line 49: The params in the function description comment currently don't match the actual params 1:1.

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+Done:
+  return Status;
+}
+
+/**
+  Validates a specific region in the target image buffer denoted by [Hdr->Offset: Hdr->Offset + Hdr->Size]
+  is not a zero buffer.
+  @param[in] TargetImage  The pointer to the target image buffer.
```
makubacki, line 94: For all of these, the sections need some whitespace separation.

```suggestion
  is not a zero buffer.

  @param[in] TargetImage  The pointer to the target image buffer.
```

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
```diff
@@ -0,0 +1,325 @@
...
+  @param[in] Hdr          The header of the validation entry.
+  @return EFI_SUCCESS             The target image passes the validation.
```
makubacki, line 96: `@retval` for individual status codes.

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
```diff
@@ -0,0 +1,325 @@
...
+**/
+EFI_STATUS
+PeCoffImageValidationNonZero (
+  IN VOID                           *TargetImage,
+  IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr
+  )
+{
+  EFI_STATUS  Status;
+
+  if (IsZeroBuffer ((UINT8 *)TargetImage + Hdr->Offset, Hdr->Size)) {
```
makubacki, line 107: `MemAttribute` is checked for `NULL` above in `InspectTargetRangeAttribute()` but the pointers dereferenced here are not. Is that intentional?

SeaPkg/Include/Library/PeCoffValidationLib.h
Javagedes: Good catch. Thanks. Updated.

SeaPkg/Include/Library/PeCoffValidationLib.h
```diff
@@ -0,0 +1,117 @@
...
+  @param[in] Hdr          The header of the validation entry.
+
+  @return EFI_SUCCESS             The target image passes the validation.
+  @return EFI_SECURITY_VIOLATION  The specified buffer in the target image is all zero.
+**/
+EFI_STATUS
+PeCoffImageValidationNonZero (
+  IN VOID                           *TargetImage,
+  IN IMAGE_VALIDATION_ENTRY_HEADER  *Hdr
+  );
```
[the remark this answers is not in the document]
Javagedes: Added. Thanks

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
```diff
@@ -0,0 +1,325 @@
...
+  )
+{
+  UINT64                     MemAttr;
+  IMAGE_VALIDATION_MEM_ATTR  *MemAttrHdr;
+  EFI_PHYSICAL_ADDRESS       AddrInTarget;
+  EFI_STATUS                 Status;
+
+  MemAttrHdr = (IMAGE_VALIDATION_MEM_ATTR *)Hdr;
+  if ((MemAttrHdr->TargetMemoryAttributeMustHave == 0) && (MemAttrHdr->TargetMemoryAttributeMustNotHave == 0)) {
+    DEBUG ((DEBUG_ERROR, "%a: Current entry 0x%p has invalid must have 0x%x and must not have 0x%x\n", __func__, MemAttrHdr, MemAttrHdr->TargetMemoryAttributeMustHave, MemAttrHdr->TargetMemoryAttributeMustNotHave));
```
makubacki: Since this is only printed if both are zero, can this be simpler and more helpful?

```suggestion
    DEBUG ((DEBUG_ERROR, "%a: Entry at 0x%p cannot have zero for must and must not have attribute values.\n", __func__, MemAttrHdr, MemAttrHdr->TargetMemoryAttributeMustHave, MemAttrHdr->TargetMemoryAttributeMustNotHave));
```

SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.inf
Javagedes: No reason. I've updated it to match the inf file name.

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
Javagedes: Fixed here and in multiple other places. Thanks!

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
```diff
@@ -0,0 +1,325 @@
...
+    Status = EFI_INVALID_PARAMETER;
+    goto Done;
+  }
+
+  if (Hdr->Size > sizeof (AddrInTarget)) {
+    DEBUG ((DEBUG_ERROR, "%a: Current entry 0x%p has invalid size: 0x%x\n", __func__, Hdr, Hdr->Size));
```
makubacki:
```suggestion
    DEBUG ((DEBUG_ERROR, "%a: Current entry 0x%p has invalid size of 0x%x, max is 0x%x\n", __func__, Hdr, Hdr->Size, sizeof (AddrInTarget)));

```

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
Javagedes: Fixed the spacing here and in multiple other places!

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
Javagedes: Ah good to know. I've updated accordingly throughout the .c and .h file

makubacki (approved):

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
Javagedes: Makes sense to me. Updated!

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
Javagedes: When all of this was mashed together, the Caller (PeCoffImageDiffValidation) did all of the sanity checks (and still does). 

Now that we have a library interface separating the two, it is probably important to add these types of checks into the individual validation functions. 

Do you have a preference on solution? (duplicating safety checks in BasePeCoffLibNegative and BasePeCoffValidationLib vs moving them from BasePeCoffLibNegative to BasePeCoffValidationLib vs Keeping as is (cxhecks soley in BasePeCoffLibNegative)

SeaPkg/Include/Library/PeCoffValidationLib.h
Javagedes: I've updated all pointer usage across all of the validation functions to be `CONST` as they are all read only in their usage. This mainly pertains to the `*TargetImage` and `*Hdr`, but also any other pointers.

SeaPkg/Library/BasePeCoffValidationLib/Validation.c
makubacki: Where I've settled on this, is to let each function validate based on its direct usage. If a function (one in `BasePeCoffNegativeLib`) is simply passing a pointer to another function (one in `BasePeCoffValidationLib`) but not actually dereferencing it, then it is okay to let the callee perform the validation (no redundancy). Though, I don't view redundancy in cases where it occurs as that concerning and arguments can be made about defensive programming which I generally agree with for how code is used and refactored in practice.

In the end, external interfaces, like what you noted with these being moved to their own library interface, need to make sure their externally provided arguments are validated within their call stack prior to usage and the result bubbled to the caller in a consistent, reasonable way so that their return status code is accurate.

Javagedes: Sounds good. I added parameter validation checks to all the library functions in the new library, and removed any duplicate checks no longer applicable from the caller.

SeaPkg/Include/Library/PeCoffValidationLib.h
```diff
@@ -0,0 +1,130 @@
...
+EFIAPI
+PeCoffImageValidationContent (
+  IN CONST VOID                           *TargetImage,
+  IN CONST IMAGE_VALIDATION_ENTRY_HEADER  *Hdr,
+  IN CONST IMAGE_VALIDATION_DATA_HEADER   *ImageValidationHdr
+  );
+
+/**
+  Validates a specific region in the target image buffer denoted by [Hdr->Offset: Hdr->Offset + Hdr->Size]
+  belongs to a user page that is mapped inside MM and teh page attributes match the requirements specified
```
kenlautner, line 66: nit: spelling.

Javagedes: Updated `teh` to `the`. Thanks :)

SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.c
```diff
@@ -0,0 +1,451 @@
+/** @file
+  Defines function definitions for validating a loaded PE/COFF image post execution
+  against an auxiliary file (defined in SeaAuxiliary.h) that contains auxiliary
```
kenlautner, line 3: I think this is supposed to say `contains multiple auxiliary validation entries`

SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.c
```diff
@@ -0,0 +1,451 @@
...
+
+/**
+  Validates a specific region in the target image buffer denoted by [Hdr->Offset: Hdr->Offset + Hdr->Size]
+  is not a zero buffer.
+
+  @param[in] TargetImage  The pointer to the target image buffer.
+  @param[in] Hdr          The header of the validation entry.
+
+  @retval EFI_SUCCESS             The target image passes the validation.
+  @retval EFI_SECURITY_VIOLATION  The specified buffer in the target image is all zero.
```
kenlautner, line 102: You should add an EFI_INVALID PARAMETER return value.

SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.c
Javagedes: Yes, thanks. Fixed:)

SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.c
```diff
@@ -0,0 +1,451 @@
...
+      Hdr,
+      OriginalImageBaseAddress
+      ));
+    Status = EFI_INVALID_PARAMETER;
+    goto Done;
+  }
+
+  if ((Hdr->EntrySignature != IMAGE_VALIDATION_ENTRY_SIGNATURE) || (Hdr->ValidationType != IMAGE_VALIDATION_ENTRY_TYPE_SELF_REF)) {
+    DEBUG ((DEBUG_ERROR, "%a: Invalid entry signature 0x%x or type 0x%x at 0x%p\n", __func__, Hdr->EntrySignature, Hdr->ValidationType, Hdr));
+    Status = EFI_COMPROMISED_DATA;
```
kenlautner, line 356: For invalid type and signature the return status is EFI_COMPROMISED_DATA here but in every other function it's EFI_INVALID_PARAMETER.  Personally I think it should be EFI_COMPROMISED_DATA in the other functions as well but at minimum it should be consistent across them all.

SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.c
```diff
@@ -0,0 +1,453 @@
...
+{
+  EFI_STATUS  Status;
+
+  if ((TargetImage == NULL) || (Hdr == NULL)) {
+    DEBUG ((DEBUG_ERROR, "%a: At least one invalid input parameter: TargetImage 0x%p, Hdr 0x%p\n", __func__, TargetImage, Hdr));
+    Status = EFI_INVALID_PARAMETER;
+    goto Done;
+  }
+
+  if ((Hdr->EntrySignature != IMAGE_VALIDATION_ENTRY_SIGNATURE) || (Hdr->EntrySignature != IMAGE_VALIDATION_ENTRY_TYPE_NON_ZERO)) {
```
kenlautner, line 121:
```suggestion
  if ((Hdr->EntrySignature != IMAGE_VALIDATION_ENTRY_SIGNATURE) || (Hdr->ValidationType != IMAGE_VALIDATION_ENTRY_TYPE_NON_ZERO)) {

```

Javagedes: 🤦 Good catch

Javagedes: Fixed, thank you :)

PR #406, opened 2025-02-22
[REBASE&FF] Add Pointer validation customization for `in_mseg`

Javagedes:
## Description

Adds a new customization to the pointer validation type that will validate if the pointer is in MSEG or not.

if `in_mseg` == `true`, then the pointer must be in MSEG
if `in_mseg` != `true`, then the pointer must *not* be in MSEG

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

<_Describe the test(s) that were run to verify the changes._>

## Integration Instructions

Consumers must update any `Pointer` rules and set `in_mseg` if the pointer is expected to be in MSEG.

```toml
validation.type = "pointer"
validation.in_mseg = true
```

### DIFF
--- a/SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
+++ b/SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
@@ -26,6 +26,7 @@
 #include <Uefi.h>
 #include <Base.h>
 #include <SeaAuxiliary.h>
+#include <Library/BaseLib.h>
 #include <Library/PeCoffLib.h>
 #include <Library/BaseMemoryLib.h>
 #include <Library/DebugLib.h>
@@ -34,6 +35,9 @@
 #include <Library/PeCoffValidationLib.h>
 #include <Library/PeCoffLibNegative.h>
 #include <IndustryStandard/PeImage.h>
+#include <Register/Intel/ArchitecturalMsr.h>
+#include <Register/Intel/StmApi.h>
+#include <Register/Intel/Cpuid.h>
 
 /**
   Retrieves the PE or TE Header from a PE/COFF or TE image.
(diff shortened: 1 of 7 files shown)


kuqin12 (approved):

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
```diff
@@ -686,6 +690,80 @@ PeCoffLoaderImageNegativeReadFromMemory (
   return RETURN_SUCCESS;
 }
 
+/**
+
```
makubacki, line 694: nit

```suggestion
```

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
```diff
@@ -686,6 +690,80 @@ PeCoffLoaderImageNegativeReadFromMemory (
...
+  EFI_STATUS                         Status;
+  MSR_IA32_SMM_MONITOR_CTL_REGISTER  SmmMonitorCtl;
+  CPUID_VERSION_INFO_EBX             VersionInfoEbx;
+  STM_HEADER                         *StmHeader;
+  UINTN                              NumberOfCpus;
+
+  //
+  // Find the MSEG Base address
+  //
+  SmmMonitorCtl.Uint64 = AsmReadMsr64 (MSR_IA32_SMM_MONITOR_CTL);
```
makubacki, line 740: I believe this should also check that `SmmMonitorCtl.Bits.Valid == 1`. It is supposed to be set whenever the STM image loader (UEFI) has loaded SEA into memory and set the MSEG base in the MSR. 

My understanding is that it would be set in `SmmCpuFeaturesLib` on the SMI rendezvous and it's required for the SEA VMCALL to succeed. So, it should be set during STM execution. However, `BasePeCoffNegativeLib` is also linked against the `ResponderValidationTestLib`. I'd expect it to be set there as well since MSEG should be loaded by shell and it shouldn't be cleared until reset to my understanding.

makubacki (approved):

PR #411, opened 2025-03-03
gen_aux: Add rule coverage report

Javagedes:
## Description

Adds an auto-generated json file containing data on sections of the `efi` image that have validation rules applied to it.

### Report Format

```json
{
  "coverable": "0x0",
  "coverd": "0x0",
  "covered_percent": "0.00%",
  "coverable_by_rules": "0x0",
  "covered_by_rules": "0x0",
  "covered_by_rules_percent": "0.00%",
  "segments": [
    {
      "symbol": "",
      "start": "0x0",
      "end": "0x0",
      "covered": true,
      "reason": "",
    }
  ]
}
```

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Ensure auxfile generation works as expected.

## Integration Instructions

N/A

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/report.rs
+++ b/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/report.rs
@@ -1,0 +1,301 @@
+//! A module for generating a report showing the the coverage of the validation rules.
+//! 
+//! ## License
+//!
+//! Copyright (c) Microsoft Corporation.
+//!
+//! SPDX-License-Identifier: BSD-2-Clause-Patent
+//!
+use anyhow::{anyhow, Result};
+use serde::Serialize;
+
+use std::io::Write;
+use std::cmp::Ordering;
+
+use crate::{auxgen::{AuxFile, ImageValidationEntryHeader}, validation::ValidationType};
+
+/// A struct representing the coverage report for the firmware image.
+/// 
+/// This struct is serialized to a JSON file for easy viewing.
+#[derive(Serialize)]
+pub struct CoverageReport {
+    /// The total number of bytes that can be considered covered. e.g. the loaded image size.
+    coverable: String,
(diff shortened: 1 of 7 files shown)


SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/report.rs
```diff
@@ -0,0 +1,321 @@
...
+                    self.insert(Segment::new(section.virtual_address, next_section_start, true, "ReadExecute Section".to_string()))?
+                }
+                _ => continue,
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Finds the padding at the end of each section and adds it as a covered segment.
```
kenlautner, line 170: I think it would make more sense to remove the padding from the overall image size instead of adding it to the section size.  This would give us a more accurate covered% as padding will increase our average.

Javagedes: This is actually one of the reasons why I provided two calculations:

1. All coverable: `(PE header + covered sections + padding + rules) / Total size`
2. Only coverable by rules: `rules / (rules) + uncovered)`

I'm completely fine with updating the final calculation for (1) to be `(PE Header + covered sections + rules) / (Total Size - padding)` If we think it is a more accurate view of the data, but I think keeping the padding segments in the raw data is valuable

kenlautner (approved):

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/report.rs
Javagedes: No other requests for change. Closing.

makubacki (approved):

kuqin12 (approved):

PR #425, opened 2025-03-17
gen_aux: raw data report updates

Javagedes:
## Description

Updates the raw data report with the following:

1. A new top level entry, sections, that contains a list of the PDB sections that are present in the PDB (and a single non-section section representing the PE Header).

2. Updates segments representing a symbol with the validation type of `ValidationType::None` to be considered covered rather than uncovered with the understanding that this covered reason is extremely weak.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated the json raw data output now contains the following:

```json
"sections": [
    {
      "name": "PE Header",
      "start": "0x0",
      "end": "0x1000"
    },
    {
      "name": ".text",
      "start": "0x1000",
      "end": "0x28000"
    },
    {
      "name": ".rdata",
      "start": "0x28000",
      "end": "0x37000"
    },
   ...
]
```

Validated that "Validation Rule: None" types are now marked as covered

```json
    {
      "symbol": "gMmCoreMmst.CurrentlyExecutingCpu",
      "start": "0x37388",
      "end": "0x37390",
      "covered": true,
      "reason": "Validation Rule: None"
    },
```
## Integration Instructions

N/A

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/report.rs
+++ b/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/report.rs
@@ -12,7 +12,7 @@ use serde::Serialize;
 use std::io::Write;
 use std::cmp::Ordering;
 
-use crate::{auxgen::{AuxFile, ImageValidationEntryHeader}, validation::ValidationType};
+use crate::auxgen::{AuxFile, ImageValidationEntryHeader};
 
 /// A struct representing the coverage report for the firmware image.
 /// 
@@ -31,6 +31,7 @@ pub struct CoverageReport {
     covered_by_rules: String,
     /// The percentage of the total number of bytes that are covered by validation rules.
     covered_by_rules_percent: String,
+    sections: Vec<Section>,
     /// A list of segments of the loaded image and their coverage status.
     segments: Vec<Segment>,
 }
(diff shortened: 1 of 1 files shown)


kuqin12 (approved):

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/report.rs
```diff
@@ -82,6 +88,58 @@ impl CoverageReport {
...
+        }
+    }
+
+    /// Consumes the list and returns the inner vector of sections.
+    pub fn into_inner(self) -> Vec<Section> {
+        self.sections
+    }
+
+    /// Adds a new section to the list with the given name, start address, and end address.
+    fn add_section(&mut self, name: String, start: u32, end: u32) {
```
makubacki, line 112: The image size as a 4-byte value makes sense because `SizeOfImage` is 4-bytes in both PE32/PE32+. But the addresses could be 64-bit, right?

makubacki (approved):

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/report.rs
Javagedes: The `VirtualAddress`, field in a Section Header of a PE image is 4-bytes. While there is a possibility that virtual address + size can overflow, that would theoritically break the next sections `VirtualAddress` field, so I think its OK.

Javagedes: All the data I'm getting from the pdb parser is u32. I'm not doing any type of `usize` to `u32` conversions either

makubacki: If this does not include the base (RVA), I agree.

PR #428, opened 2025-03-18
Fixed issues with the pointer validation code

kenlautner:
## Description

The pointer validation entry type has some issues with the validation logic.
1. It doesn't correctly increment with the entry list iteration leading to us point to nonvalid memory.
2. The calculation for the MSEG region was incorrect and was providing a random address for BASE and the incorrect number of CPU cores.
3. The target address was incorrectly being set to the location of symbol in the MM Supervisor image.  It has been updated to use the address of pointer so we can correctly judge if it's in or outside MSEG.

This changes address both of these issues by casting the entry type when incrementing to address (1) and then using the correct data structures to calculate the MSEG region.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Intel physical platform.  The validation process correctly gets MSEG.

## Integration Instructions

N/A

### DIFF
--- a/SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
+++ b/SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
@@ -38,6 +38,7 @@
 #include <Register/Intel/ArchitecturalMsr.h>
 #include <Register/Intel/StmApi.h>
 #include <Register/Intel/Cpuid.h>
+#include <x64/Vmx.h>
 
 /**
   Retrieves the PE or TE Header from a PE/COFF or TE image.
@@ -728,9 +729,10 @@ GetMsegBaseAndSize (
 {
   EFI_STATUS                         Status;
   MSR_IA32_SMM_MONITOR_CTL_REGISTER  SmmMonitorCtl;
-  CPUID_VERSION_INFO_EBX             VersionInfoEbx;
+  CPUID_CACHE_PARAMS_EAX             CacheParamsEax;
   STM_HEADER                         *StmHeader;
   UINTN                              NumberOfCpus;
+  UINT32                             MaxStandardCpuIdIndex;
 
   //
   // Find the MSEG Base address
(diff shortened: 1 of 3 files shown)


SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
```diff
@@ -749,14 +751,21 @@ GetMsegBaseAndSize (
...
-  AsmCpuid (CPUID_VERSION_INFO, NULL, &VersionInfoEbx.Uint32, NULL, NULL);
-  NumberOfCpus = VersionInfoEbx.Bits.MaximumAddressableIdsForLogicalProcessors;
+  StmHeader = (STM_HEADER *)(UINTN)((UINT32)AsmReadMsr64 (IA32_SMM_MONITOR_CTL_MSR_INDEX) & 0xFFFFF000);
+
+  AsmCpuid (CPUID_SIGNATURE, &MaxStandardCpuIdIndex, NULL, NULL, NULL);
+  if (MaxStandardCpuIdIndex >= CPUID_CACHE_PARAMS) {
+    AsmCpuidEx (CPUID_CACHE_PARAMS, 0, &CacheParamsEax.Uint32, NULL, NULL, NULL);
+
+    if (CacheParamsEax.Uint32 != 0) {
+      NumberOfCpus = CacheParamsEax.Bits.MaximumAddressableIdsForLogicalProcessors + 1;
```
kenlautner, line 766: This is definitely the correct way to query cpu cores for intel hardware but the value still doesn't quite make sense.  If anyone has an knowledge on the subject please share.

kenlautner: FYI it returns 2 currently when I would expect 8 on my device.

kuqin12: What about `MaximumAddressableIdsForProcessorCores`?

makubacki: - Is there a reason to not at least first consider using `gMpInformation2HobGuid`?

- CPU hardware logic should not be directly implemented in `BasePeCoffLibNegative`, it should come from a CPU interface.

---

`MaximumAddressableIdsForLogicalProcessors` is the number of logical processors sharing a cache. This is intended for code to take advantage of identifying things like performance efficiencies that can be gained by taking advantage of the cache at a given cache hierarchy level. It's not what you want which is a generic core count.

`MaximumAddressableIdsForProcessorCores` would be the number of processor cores in the physical package which is closer to what you want for client systems.

However, as mentioned, this really should be abstracted from this library using a CPU interface. Often, an MSR is used to get this information under those interfaces.

kuqin12: This will be run in the STM context. I do not think HOBs are available (or should be referenced) in this case. But CPU interface is a reasonable request of change. Do you happen to know which MSR that is, if publicly available?

makubacki: The HOBs don't need to be referenced directly, but they are a reliable source of platform abstracted information. It is not publicly available to my knowledge.

SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.c
```diff
@@ -460,7 +461,9 @@ PeCoffImageValidationPointer (
     goto Done;
   }
 
-  InMseg = ((UINTN)((UINT8 *)TargetImage + Hdr->Offset) >= MsegBase) && ((UINTN)((UINT8 *)TargetImage + Hdr->Offset) < MsegBase + MsegSize);
+  AddrInTarget = 0;
+  CopyMem (&AddrInTarget, (UINT8 *)TargetImage + Hdr->Offset, Hdr->Size);
```
makubacki, line 465: Is this meant to copy `Hdr->Size` bytes? `AddrInTarget` is a `EFI_PHYSICAL_ADDRESS`. This code would also need to check the value of `Hdr->Size`...

I may be misunderstanding but it seems like you want to set `AddrInTarget` to the address of that offset?

```suggestion
  AddrInTarget = ((EFI_PHYSICAL_ADDRESS)(UINTN)((UINT8 *)TargetImage + Hdr->Offset));
```

makubacki: Clarified offline that this is the resolved address in MSEG at this location and the size of the content is verified earlier against the Hdr->Size.

kuqin12 (dismissed):

makubacki (dismissed):

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
kenlautner: Updated to use the STM_HEADER to store the cpu core count in the new CPU_INFORMATION_HEADER.

kuqin12 (approved):

apop5 (approved):

PR #433, opened 2025-03-21
rule coverage: Add helper function to validate

Javagedes:
## Description

Add a helper function that parses the raw rule coverage data and will return a true or false if the coverage percentage meets the platform's expectations.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated it returns true or false depending on if it meets expectations or not.

## Integration Instructions

Consumers can use `self.Helper.validate_rule_coverage(...)`

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
+++ b/SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
@@ -242,7 +242,7 @@ class GenSeaArtifacts(IUefiHelperPlugin):
             aux_path: Path to the auxiliary file.
         
         Raises:
-            FileNotFoundError: The auxiliary file does not exist.
+            FileNotFoundError: The auxiliary file path does not exist or is not a file.
             json.JSONDecodeError: The auxiliary file is not a valid JSON file.
 
         Returns:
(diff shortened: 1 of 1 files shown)


SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
```diff
@@ -231,6 +233,57 @@ def generate_sea_manifest(stm_bin: Path, output_path: Path, config: dict, worksp
         
         return 0
 
+    @staticmethod
+    def validate_rule_coverage(percent: int, aux_path: Path):
```
makubacki, line 237:
```suggestion
    def validate_rule_coverage(percent: int, aux_path: Path) -> bool:
```

makubacki: Was `percent` meant to be an `int` instead of a `float`?

makubacki (approved):

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
Javagedes: Yes I did int purposefully, but I'd say you are right, a float is better. I'll update.

Javagedes: Updated.

kuqin12 (approved):

PR #437, opened 2025-03-27
aux_gen: Add support for symbols that are lists

Javagedes:
## Description

previously, lists were treated as any other symbol, meaning a rule applied to the list itself instead of individual elements of the list. This PR updates the functionality to treat each element in the list as an individual symbol. If a rule is applied to the list, it is actually treated as if you are applying the rule to the underlying type, and a validation entry header is generated for each entry in the list. 

This pull request comes with some new configuration values in the `[[rule]]` config entry:

1. `index`, which tells the tool to only generate this particular rule for the specified index. If the index is larger than the actual length, the tool stops with an error
2. `sentinel`, which tells the tool to generate a special `Content` rule for the last entry in the list. This rule requires that the value must be zero.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested the following scenarios:

1. Specifying an index too large results in the tool stopping with a properly defined error
2. Specifying a valid index only applies the rule to that index
3. Specifying sentinel = true correctly turns the rule into a Content rule
4. Specifying `field` on a list symbol correctly calculates the offset and size for that field, for each entry in the list
5. Manual specification of `offset` continues to work in lieu of `field`
 
## Integration Instructions

Config file maintainers must validate any rules that are applied to a list also makes sense to be applied to individual elements in said list.


kuqin12:
Do you mean "list" as linked list in the original C source code of the supv? Or some other structure from the pdb crate we used?

Javagedes:
> Do you mean "list" as linked list in the original C source code of the supv? Or some other structure from the pdb crate we used?

@kuqin12 I mean arrays. Specifically arrays that are statically sized and not on the heap.

I'll update my PR :)

SeaPkg/Tools/GenSeaArtifacts/gen_aux/readme.md
```diff
@@ -62,6 +65,8 @@ validation.type = 'Required[String]'
 - `scope`: If specified, the rule is only applied when this scope is active. Otherwise it is always applied.
 - `symbol`: Determines the address and size for the rule
 - `field`: Updates the address and size to be that of the field, rather than the symbol itself.
+- `index`: If the symbol is a list, only apply the rule to the specified index.
+- `sentinel`: If the symbol is a list, apply content rule to only the final rule such that it's content must be all zeros
```
makubacki, line 69:
```suggestion
- `sentinel`: If the symbol is a list, apply content rule to only the final rule such that its content must be all zeros
```

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/auxgen.rs
```diff
@@ -171,18 +178,53 @@ impl <'a> ctx::TryIntoCtx<Endian> for &ImageValidationEntryHeader {
...
-        entry.validation_type = rule.validation.clone();
-        if let Some(ref field) = rule.field {
-            entry.symbol = format!("{}.{}", &rule.symbol, field);
-        } else {
-            entry.symbol = rule.symbol.clone();
+    /// Generates one or more ImageValidationEntryHeaders from a ValidationRule and a Symbol.
+    ///
+    /// Multiple headers are generated if the top level symbol is a list of an underlying type.
+    /// In this scenario, the same rule is applied to each element in the list rather than the
+    /// list as a whole.
```
makubacki, line 185: Similar to any earlier comment in the PR, I'd drop "list" here and replace it either with "array" (preferred since this is C and that's clear per the static arrangement of contents in memory - fixed size and random access) or "sequenced type" (less preferred since it's unnecessarily ambiguous). "list" will be associated with linked lists for a lot of people in C.

Javagedes: This makes sense. Updated all references to list to be "array"

makubacki:
@javagedes, what is the unit test status?

Javagedes:
> @Javagedes, what is the unit test status?

@makubacki there are unfortunately zero unit tests for this tool. It's all been testing by hand. Once we get the needed functionality out for code complete, I'll go back through and write unit tests. I can do it now if needed but it will take a fair amount of time.

makubacki:
> > @Javagedes, what is the unit test status?
> 
> @makubacki there are unfortunately zero unit tests for this tool. It's all been testing by hand. Once we get the needed functionality out for code complete, I'll go back through and write unit tests. I can do it now if needed but it will take a fair amount of time.

No problem. I'm fine for them to go in later and not let it hold back this PR.

makubacki (approved):

SeaPkg/Tools/GenSeaArtifacts/gen_aux/readme.md
```diff
@@ -46,14 +46,17 @@ offset = 'Optional[u32]'
 
 ### rule
```
kenlautner, line 47: Leaving a comment as a note to hold off on completing this PR.  I want to get my PR(s) in before this so that additional change isn't required (for now) for my config file PR.

kenlautner (approved):

PR #438, opened 2025-03-28
cov_report: Add symbol names to segments missing coverage

Javagedes:
## Description

Updates the coverage report generation to add symbol names to any segments that are missing coverage, if a symbol name exists. If multiple symbols exist for a single segment, the segment is split accordingly. symbols with only partial coverage (due to a rule using the offset / field config option) will now also show the symbol name for any portion of the symbol that is not covered.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated the following scenarios:

1. symbols names are added to uncovered sections
2. symbols that are partially covered due to a field rule will have all other parts of the symbol that are uncovered properly marked with the symbol name
3. a single segment marked as uncovered that contains multiple symbols properly shows the multiple symbols.

## Integration Instructions

N/A

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/main.rs
+++ b/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/main.rs
@@ -188,7 +188,8 @@ pub fn main() -> Result<()> {
             }
         }
 
-        let report = CoverageReport::build(&efi, &aux)?;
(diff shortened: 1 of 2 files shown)


kuqin12 (approved):

makubacki (approved):

Javagedes:
This relies on some changes in #437

PR #447, opened 2025-04-03
aux_gen: Auto generate rules for padding

Javagedes:
## Description

While developing the coverage report raw data, it was noted that some spaces between symbols in the .data section were not associated with any symbols in the PDB. It was determined that this was padding in the symbols themselves which work to align them in the file. This pull request does the following things:

1. Cleans up the final build into multiple separate functions
2. Generates a content validation rule (of all zero) for any byte ranges not assocaited with a symbol
3. Generates a content validation rule (of all zeros) for any padding between sections, that is in a section that is both Read and Write.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

<_Describe the test(s) that were run to verify the changes._>

## Integration Instructions

<_Describe how these changes should be integrated. Use N/A if nothing is required._>

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/auxgen.rs
+++ b/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/auxgen.rs
@@ -10,10 +10,10 @@
 use std::{fmt, io::Write, mem::size_of};
 use pdb::{SectionCharacteristics, TypeInformation};
 use scroll::{self, ctx, Endian, Pread, Pwrite, LE};
-use crate::{type_info::TypeInfo, ConfigFile, KeySymbol, ValidationRule, ValidationType};
+use crate::{report::CoverageReport, type_info::TypeInfo, ConfigFile, KeySymbol, ValidationRule, ValidationType};
 
 /// The type of symbol in the PDB file.
-#[derive(Default, Clone, PartialEq)]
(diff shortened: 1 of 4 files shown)


kuqin12 (approved):

makubacki (approved):

kuqin12 (approved):

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/auxgen.rs
```diff
@@ -409,13 +417,83 @@ impl AuxBuilder {
...
+                    String::from("Section Padding")
+                } else {
+                    String::from("Symbol Padding")
+                };
+
+                rule.symbol_info = Some(Symbol {
+                    name,
+                    address: segment.start() as u32,
+                    type_info: TypeInfo::one(size as u64, None),
+                    symbol_type: SymbolType::None,
```
kenlautner, line 475: `section_characteristics: SectionCharacteristics::default()` needs to be added to be compatible with the latest changes.

Javagedes: Added as suggested. Thank you!

kenlautner (approved):

PR #452, opened 2025-04-07
Remove/Modify some global variables

kenlautner:
## Description

Some global variables exist that are unnecessary because they're values are constant or they're only used in testing scenarios.  This change removes the unnecessary variables and adds guards around the globals that are used for testing so they only are in builds which enable the testing PCD.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Physical platform.  Behavior remains the same with or without this change.

## Integration Instructions

N/A

### DIFF
--- a/MmSupervisorPkg/Core/FwVol/FwVol.c
+++ b/MmSupervisorPkg/Core/FwVol/FwVol.c
@@ -111,6 +111,7 @@ Returns:
       TotalSize = TotalSize + FileSize;
     }
   } while (!EFI_ERROR (Status));
+
   // If by the time we get here this FV is outside of MMRAM, copy it MMRAM
   // It will be marked as CPL3 RO XP before entering MMI
   Status = MmAllocatePages (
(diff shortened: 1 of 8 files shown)


MmSupervisorPkg/Core/FwVol/FwVol.c
```diff
@@ -23,19 +23,6 @@ typedef struct {
   EFI_MEMORY_DESCRIPTOR    DriverCacheDesc;
 } FFS_DRIVER_CACHE_LIST;
 
-//
-// List of file types supported by dispatcher
-//
-EFI_FV_FILETYPE  mMmFileTypes[] = {
```
Javagedes, line 29: You might be able to get away with just adding CONST here

Javagedes:
I think for some of these you might be able to just add the CONST identifier to get the symbols moved from the .data section to the .rdata section

kuqin12 (approved):

Javagedes (approved):

PR #455, opened 2025-04-09
Override VS2022 release DLINK_FLAGS for the MmSupervisorCore

kenlautner:
## Description

Override VS2022 release DLINK_FLAGS for the MmSupervisorCore.  The only change is that the .data and .rdata sections are no longer merged which is necessary for SEA and the aux generation script.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on intel physical platforms.  Release builds have the expected coverage with this change.

## Integration Instructions

N/A


Javagedes (approved):

kuqin12 (approved):

MmSupervisorPkg/Core/MmSupervisorCore.inf
```diff
@@ -240,6 +240,8 @@
 [BuildOptions.common]
   #Subsystem version will be used as MmSupervisor driver version
   #Image version will be used as MmSupervisor security patch level(SPL) value
+  #Override the VS2022 RELEASE DLINK_FLAGS so the .data and .rdata sections are not merged
+  RELEASE_VS2022_X64_DLINK_FLAGS == /NOLOGO /NODEFAULTLIB /IGNORE:4001 /IGNORE:4281 /IGNORE:4254 /OPT:REF /OPT:ICF=10 /MAP /ALIGN:32 /SECTION:.xdata,D /SECTION:.pdata,D /Machine:X64 /LTCG /DLL /ENTRY:$(IMAGE_ENTRY_POINT) /SAFESEH:NO /BASE:0 /DRIVER /ALIGN:4096 /DLL
```
makubacki, line 244: This is not ideal for keeping in sync with the original tools_def RELEASE values and keeping consistent with common values in the DEBUG build that would come from there.

kenlautner: Is there a cleaner method for preventing the merge of the .rdata and .data sections?

makubacki: Can you try this https://github.com/microsoft/mu_basecore/pull/1328? Edit your local Config/tools_def.txt file to have the same changes in the template there.

makubacki: It looks like that won't work and we could potentially do something with environment variables but I think what you're doing here is fine for now.

Do you think (1) This could apply to all VS not just VS2022 so it doesn't subtly not work if another VS version is used (2) You could also copy & paste the DEBUG DLINK flags so at least they're both defined here so they're consistent in their definition location?

kenlautner: Got it.  I've added the DEBUG DLINK flags to also be overridden for consistency.

kuqin12 (approved):

makubacki (approved):

PR #458, opened 2025-04-11
aux_gen: properly calculate array length

Javagedes:
## Description

A bug was introduced in the code that calculates the total element count of an array because it did not consider the case for when the symbol is a multi-dimensional array. This is because the `total_size` function returns the total size of the element in the array, which when working with a multi-dimensional array, is the size of the row (array), instead of the underlying element.

To resolve this issue, we instead calculate the total element count by using the `element_size()` function, which returns the size of the underlying type instead. Due to this recursive nature of `TypeInfo::from_type_index`, this is ultimately the actual element type, even if the symbol is a 2, 3, 4, etc. dimensional array.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated a symbol that is a 2-Demensional array now has entries generated equal to MxN.

## Integration Instructions

N/A

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/type_info.rs
+++ b/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/type_info.rs
@@ -135,7 +135,8 @@ impl TypeInfo {
             TypeData::Array(arr) => {
                 let total_size = arr.dimensions[0] as u64;
                 let element = TypeInfo::from_type_index(info, arr.element_type)?;
-                TypeInfo::many(element.element_size(), total_size / element.element_size(), element.type_id())            }
+                TypeInfo::many(element.element_size(), total_size / element.element_size(), element.type_id())
+            }
             pdb::TypeData::Union(union) => {
                 TypeInfo::one(union.size, Some(index))
             }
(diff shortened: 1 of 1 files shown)


kenlautner (approved):

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/type_info.rs
```diff
@@ -135,7 +135,7 @@ impl TypeInfo {
             TypeData::Array(arr) => {
                 let total_size = arr.dimensions[0] as u64;
                 let element = TypeInfo::from_type_index(info, arr.element_type)?;
-                TypeInfo::many(element.total_size(), total_size / element.total_size(), element.type_id())
+                TypeInfo::many(element.element_size(), total_size / element.element_size(), element.type_id())
```
makubacki, line 138: The PR description might be valuable to summarize here to clarify how multi-dimensional arrays are handled.

makubacki (approved):

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/type_info.rs
Javagedes: Added some documentation to the update. Good thought.

kenlautner (approved):

PR #461, opened 2025-04-15
Update self ref validation to use the original image for validation

kenlautner:
## Description

Update self ref validation to use the original image for validation.  This avoids the issue where the relocation code changes the target image contents that we're trying to validate leading to a failure in the rules validation.  Instead we use the original image to validate the pointer.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on an Intel Physical platform.  The previously failing rules now pass.

## Integration Instructions

N/A

### DIFF
--- a/SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
+++ b/SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
@@ -884,7 +884,7 @@ PeCoffImageDiffValidation (
         NextImageValidationEntryHdr = (IMAGE_VALIDATION_ENTRY_HEADER *)((IMAGE_VALIDATION_MEM_ATTR *)ImageValidationEntryHdr + 1);
         break;
       case IMAGE_VALIDATION_ENTRY_TYPE_SELF_REF:
-        Status                      = PeCoffImageValidationSelfRef (OriginalImageBaseAddress, ImageValidationEntryHdr, OriginalImageBaseAddress);
+        Status                      = PeCoffImageValidationSelfRef (OriginalImageBaseAddress, ImageValidationEntryHdr);
         NextImageValidationEntryHdr = (IMAGE_VALIDATION_ENTRY_HEADER *)((IMAGE_VALIDATION_SELF_REF *)ImageValidationEntryHdr + 1);
         break;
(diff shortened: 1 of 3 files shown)


SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
```diff
@@ -884,7 +884,7 @@ PeCoffImageDiffValidation (
         NextImageValidationEntryHdr = (IMAGE_VALIDATION_ENTRY_HEADER *)((IMAGE_VALIDATION_MEM_ATTR *)ImageValidationEntryHdr + 1);
         break;
       case IMAGE_VALIDATION_ENTRY_TYPE_SELF_REF:
-        Status                      = PeCoffImageValidationSelfRef (TargetImage, ImageValidationEntryHdr, OriginalImageBaseAddress);
+        Status                      = PeCoffImageValidationSelfRef (OriginalImageBaseAddress, ImageValidationEntryHdr, OriginalImageBaseAddress);
```
makubacki, line 887: Does this API need to be updated to only accept a single image base address argument?

Javagedes (approved):

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
Javagedes: I would be OK with this

makubacki (approved):
Approving, but please clean up the API if there's not a need for it to have two parameters for the image base.

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
kenlautner: Probably a good idea.  I'll make that update.

kuqin12 (approved):

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
kuqin12: You will need to update the header file as well.

Javagedes (approved):

PR #465, opened 2025-04-18
gen_aux: Large Clean-up for reusability

### LINKED ISSUE #464: [Feature]: Automatically cover padding in symbols
### Feature Overview

If a symbol is not properly aligned, padding will be inserted in between fields in the symbol. Coverage rules are already added to cover padding between sections and between symbols themselves. Lets investigate automatically adding `Zeros` or `Content` rules for padding inbetween fields on symbols.

### Solution Overview

Option 1: For any symbol missing some portion of coverage, check to see if all fields are covered. If so, automatically insert zeros rule where needed.

Option 2: ?

### Alternatives Considered

_No response_

### Urgency

Low

I will implement the feature

No maintainer feedback needed

### Anything else?

_No response_

Javagedes:
## Description

Now that the tool is where it needs to be to provide 100% coverage for validating read / write sections of an efi image, it is time to clean up the gen_aux tool. This clean up focuses on three things: (1) Removing dead functionality that was originally added to support workarounds during development, (2) add re-usability for additional CLI tools, and (3) simplify some of the Config rule entries to reduce duplication.

### Remove dead functionality

1. Remove `excluded_symbols` as we require 100% coverage, so this no longer makes sense
2. Remove `autogen` configuration as you can now use `create-config` command line tool
3. Remove `size` and `offset` as we now properly find all necessary symbols and can insert 0 padding between symbols and fields.

### Add re-usability

This change separates the binary into a library 4 main modules, and splits the binary into two separate executables:

1. `config` which contains definitions / implementations for the TOML file
2. `file` which contains definitions / implementations for generating the auxiliary file
3. `metadata` which contains pdb data and is used to convert `config` values to `file` values
4. `report` for generating the json report

`create-aux` tool, which creates the auxiliary file from the config and pdb
`create-config` tool, which creates a configuration file that provides 100% coverage of all symbols, but all rules are `none` rules
`test-aux` tool, which is not implemented, but will test a config file against a dumped executed efi image

### Simplify config rule entries

1. Updates `array.index` to support either a single index value, or an inclusive range [a, b] for generating rules.
2. Properly calculates padding between fields in symbols and creates a 0 content rule for them.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated the tool continues to generate 100% covered auxiliary files and that the reversion and validation process passes.

## Integration Instructions

1. delete usage of `excluded_symbols`
3. delete usage of `autogen`
4. delete usage of `validation.size` and `validation.offset`. This should no longer be needed as we can now automatically create all padding rules
5. Update usage of `index` to be a range where necessary


Javagedes:
moving to draft as it failed validaiton.

Javagedes:
closes #464

kenlautner (approved):

kuqin12 (approved):

Javagedes:
@microsoft-github-policy-service agree

Javagedes:
@microsoft-github-policy-service rerun

PR #471, opened 2025-04-30
Add SEA CPU information library

kenlautner:
## Description

Create a new library called SeaCpuInformationLib that can provide CPU specific information.  Currently the only function is GetPlatformCoreCount but other functions can be added as necessary.  Platforms will need to implement their own non-NULL library and use it for their platform builds for correct CPU information to be propagated.

This PR also removes references to CpuInfoHdr which was removed from MU_BASECORE.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on an Intel physical platform.  The library interface works as intended.

## Integration Instructions

Add `SeaCpuInformationLib|SeaPkg/Library/BaseSeaCpuInformationLibNull/BaseSeaCpuInformationLibNull.inf` to your DSC if you're using PeCoffLibNegative.

### DIFF
--- a/SeaPkg/Include/Library/PlatformCpuInformationLib.h
+++ b/SeaPkg/Include/Library/PlatformCpuInformationLib.h
@@ -1,0 +1,24 @@
+/** @file
+  Platform CPU Information library header file
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef _PLATFORM_CPU_INFORMATION_LIB_
+#define _PLATFORM_CPU_INFORMATION_LIB_
+
+/**
+  Return the Platform CPU core count.
+
+  @retval The number of CPU cores on the system
+
+**/
+UINTN
+EFIAPI
+GetPlatformCoreCount (
+  VOID
+  );
+
+#endif
--- a/SeaPkg/Include/Library/PlatformCpuInformationLib.h
+++ b/SeaPkg/Include/Library/PlatformCpuInformationLib.h
@@ -6,8 +6,8 @@
 
 **/
 
-#ifndef _PLATFORM_CPU_INFORMATION_LIB_
-#define _PLATFORM_CPU_INFORMATION_LIB_
+#ifndef PLATFORM_CPU_INFORMATION_LIB_
+#define PLATFORM_CPU_INFORMATION_LIB_
 
 /**
   Return the Platform CPU core count.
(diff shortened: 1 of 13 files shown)


Javagedes (approved):

SeaPkg/Include/Library/PlatformCpuInformationLib.h
```diff
@@ -0,0 +1,24 @@
+/** @file
+  Platform CPU Information library header file
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef _PLATFORM_CPU_INFORMATION_LIB_
+#define _PLATFORM_CPU_INFORMATION_LIB_
```
makubacki, line 10:
```suggestion
#ifndef PLATFORM_CPU_INFORMATION_LIB_
#define PLATFORM_CPU_INFORMATION_LIB_
```

SeaPkg/Library/BasePlatformCpuInformationLibNull/BasePlatformCpuInformationLibNull.c
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+#include <Uefi.h>
+#include <Library/PlatformCpuInformationLib.h>
+
+/**
+  Return the Platform CPU core count.  This is the NULL implementation which will
+  always return 0.
+
+  @retval The number of CPU cores on the system
```
makubacki, line 16:
```suggestion
  @return The number of CPU cores on the system
```

kuqin12 (approved):

SeaPkg/SeaPkg.dec
```diff
@@ -31,6 +31,7 @@
   HashLibRaw|Include/Library/HashLibRaw.h
   PeCoffLibNegative|Include/Library/PeCoffLibNegative.h
   PeCoffValidationLib|Include/Library/PeCoffValidationLib.h
+  PlatformCpuInformationLib|Include/Library/PlatformCpuInformationLib.h
```
makubacki, line 34: This is a fairly generic library class name currently serving a specific use case. We might consider something like "SeaCpuInformationLib". Then if it's less likely to conflict and more obvious in build reports and DSC files that this is something specifically meant for SEA.

makubacki (approved):

Javagedes (approved):

PR #474, opened 2025-05-09
aux_gen: Add new `guid` rule to `[[rule]]`

Javagedes:
## Description

Adds a new rule, `guid`, that generates a `content` validation entry header in the aux file.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated that the generated aux file is the same exact file when replacing a content rule with the guid rule.

## Integration Instructions

Consumers can now use `validation.type = "guid"` with a `guild` field of [a, b, c, [d0, d1, d2, d3, d4, d5, d6, d7]

```toml
[[rule]]
symbol = "MyGuid"
validation.type = "guid"
validation.content = [0x0, 0x0, 0x0, [0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0]]
```

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/config.rs
+++ b/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/config.rs
@@ -11,6 +11,7 @@
 use std::ops::RangeInclusive;
 
 use serde::{Deserialize, Serialize};
+use r_efi::efi::Guid;
 
 /// The configuration file for generating an auxiliary file.
 #[derive(Debug, Default, Serialize, Deserialize)]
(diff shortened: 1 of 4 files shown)


SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/config.rs
makubacki, line 1: Is this one you could add unit tests for? Endianness has been an occasional issue with GUIDs in the past.

PR #477, opened 2025-05-13
[REBASE&FF] Add Host-based Aux validation

Javagedes:
## Description

This commit chain adds a new SEA build artifact (located in the misc/* dir) called `test-aux.exe`, which is a way to quickly generate an aux file and test it against a post-executed MM supervisor core image that is dumped from the run log. This binary links against the compiled PeCoffValidationLib object to do the actual validation, but implements some necessary functionality.

To support this, PeCoffValidationLib's `PeCoffImageValidationSelfRef` had to be updated to do pointer comparisons via the RVA rather than the actual address. Additionally, a few changes needed to be made to `aux_gen` tooling to support this.

This PR only implements three of the 5 (3/5) validation tests currently supported at runtime. The supported validation types are "SelfRef", "Content", and "NonZero". The unsupported validation tests to be added in the future are "Pointer" and "MemAttr"

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.c
```diff
@@ -414,7 +414,22 @@ PeCoffImageValidationContent (
   }
 
   if (CompareMem ((UINT8 *)TargetImage + Hdr->Offset, ContentHdr->TargetContent, Hdr->Size) != 0) {
-    DEBUG ((DEBUG_ERROR, "%a: Current entry range 0x%p: 0x%x does not match input content at 0x%p\n", __func__, ContentHdr, Hdr->Size, ContentHdr->TargetContent));
+    DEBUG ((DEBUG_ERROR, "%a: Current entry 0x%p: Content mismatch in image at address 0x%p\n", __func__, ContentHdr, ContentHdr->TargetContent));
+
+    DEBUG ((DEBUG_ERROR, "%a: Expected content: ", __func__));
```
kuqin12, line 419: Would it be better to use DUMP_HEX here?

kuqin12 (approved):

SeaPkg/Library/BasePeCoffValidationLib/BasePeCoffValidationLib.c
Javagedes: DUMP_HEX comes with some additional bloat I did not like for the purpose of showing a just a few bytes. Mainly it includes the current line you are on, and the ascii display, which I did not like.

The line looks like this:

`13:52:40.030 :     00000000: 4D 5A 00 00 00 00 00 00-00 00 00 00 00 00 00 00  *MZ..............*`

I'd be happy to switch it if you prefer.

makubacki: I have a slight preference for reusiing DUMP_HEX, but I think either could work here.

Javagedes: I'm going to keep it as is unless @kuqin12 has a strong preference. I will leave this comment open, for him to close, or respond back telling me to change it :)

PR #489, opened 2025-06-11
MmSupervisorPkg: MmSupervisorCore consume PcdMigrateTemporaryRamFirmwareVolumes.

apop5:
## Description

A platform has the ability to set PcdMigrateTemporaryRamFirmwareVolumes which tells the PEI core to migrate Firmware volumes from the flash part into system memory after gEfiPeiMemoryDiscoveredPpiGuid is installed.

When this happens, the platform may decide to leave the original EFI_HOB_TYPE_FVs in place, and expect consumers to switch to using EFI_HOB_TYPE_FV2.

Add the capability to consume EFI_HOB_TYPE_FV2 when this PCD is set. Also update the Hob consuming code to ensure that the length field of Length field of EFI_HOB_FIRMWARE_VOLUME/EFI_HOB_FIRMWARE_VOLUME2 is non zero before attempting to access/consume data from the hob.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Testing on a platform which set PcdMigrateTemporaryRamFirmwareVolumes to True.

## Integration Instructions
No integration necessary.

### DIFF
--- a/MmSupervisorPkg/Core/MmSupervisorCore.c
+++ b/MmSupervisorPkg/Core/MmSupervisorCore.c
@@ -21,6 +21,7 @@
 
 #include <Guid/MmCommonRegion.h>
 #include <Library/SecurePolicyLib.h>
+#include <Library/HobPrintLib.h>
 
 EFI_STATUS
 MmCoreFfsFindMmDriver (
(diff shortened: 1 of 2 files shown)


kuqin12 (approved):

MmSupervisorPkg/Core/MmSupervisorCore.c
```diff
@@ -711,23 +712,41 @@ DiscoverStandaloneMmDriversInFvHobs (
...
+
   Hob.Raw = GetHobList ();
   if (Hob.Raw == NULL) {
     return EFI_NOT_FOUND;
   }
 
+  // If PcdMigrateTemporaryRamFirmwareVolumes is set, the
+  // migration of temporary RAM firmware volumes can end up
+  // unmapping the fvs, so use Fv2 hobs.
+  if (FixedPcdGetBool (PcdMigrateTemporaryRamFirmwareVolumes)) {
```
makubacki, line 725: This is a platform bug. They should not pass any HOBs that are not mapped regardless of whether firmware volumes are migrated (they should always be on Intel platforms for security and perf reasons).

This PCD value is also in a separate build from the platform for SEA builds where the supervisor core is delivered as a binary so the `FixedAtBuild` value is irrelevant to the platform's setting.

MmSupervisorPkg/Core/MmSupervisorCore.c
```diff
@@ -711,23 +712,41 @@ DiscoverStandaloneMmDriversInFvHobs (
...
+    HobType = EFI_HOB_TYPE_FV2;
+  } else {
+    HobType = EFI_HOB_TYPE_FV;
+  }
+
   do {
-    Hob.Raw = GetNextHob (EFI_HOB_TYPE_FV, Hob.Raw);
+    Hob.Raw = GetNextHob (HobType, Hob.Raw);
```
makubacki, line 732: This should consider both HOB types though.

apop5:
closing. Going a different path.

PR #520, opened 2025-08-28
MmSupervisorPkg: Add MmSupervisorCoreInitLib

makubacki:
## Description

Adds a new library class that can be used to add initialization points into the MM Supervisor Core. Right now, this library just provides an opportunity to run initialization code at the very beginning of MM Supervisor Core execution. A null library instance is provided in `MmSupervisorPkg`.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Build and boot with the null lib
- Build and boot with a library instance that adds custom functionality

## Integration Instructions

- Add the `MmSupervisorPkg/Library/BaseMmSupervisorCoreInitLibNull/BaseMmSupervisorCoreInitLibNull.inf` as shown in PlatformIntegrationSteps.md to the platform DSC file.


makubacki:
This is currently added so a platform can perform serial port initialization prior to `ProcessLibraryConstructorList()` being called `MmSupervisorMain()`. Debug messages are used prior to library constructors being processed, before the serial port has an opportunity to be initialized. This could be addressed in the serial port library instance but (1) it's reasonable to for the library instance to put its initialization logic in the constructor and (2) it is likely other early init may be needed at this point.

Javagedes (approved):
Does this negate @kuqin12's plan to create a single MmSupervisorCore that is usable for all CPU architectures? Seems like this will make it different per CPU architecture.

I have no problem with it, just checking :)

apop5 (approved):

makubacki:
> Does this negate @kuqin12's plan to create a single MmSupervisorCore that is usable for all CPU architectures? Seems like this will make it different per CPU architecture.
> 
> I have no problem with it, just checking :)

It does not. The primary barrier for that will be in existing libraries linked to the core.

makubacki:
> I can foresee misunderstanding of this library, and someone introducing a library instance which has dependencies that will not be met because of the lack of services being available at this point.

I intentionally did not include an error code because those dependencies should not be introduced. I'll update the function documentation to explicitly state that.

makubacki:
> > I can foresee misunderstanding of this library, and someone introducing a library instance which has dependencies that will not be met because of the lack of services being available at this point.
> 
> I intentionally did not include an error code because those dependencies should not be introduced. I'll update the function documentation to explicitly state that.

Done

apop5 (approved):

MmSupervisorPkg/Library/BaseMmSupervisorCoreInitLibNull/BaseMmSupervisorCoreInitLibNull.c
```diff
@@ -0,0 +1,32 @@
...
+#include <Uefi.h>
+#include <Library/MmSupervisorCoreInitLib.h>
+
+/**
+  Perform any custom initialization needed by the MM Supervisor Core.
+
+  This function is called at the beginning of the MM Supervisor Core entry point,
+  before any other MM Supervisor Core initialization is performed.
+
+  This function allows any critical initialization specific to a given platform to b
```
apop5, line 19:
```suggestion
  This function allows any critical initialization specific to a given platform to be

```

Javagedes (approved):

PR #559, opened 2025-11-05
test-aux: Make it clear we are skipping some tests

Javagedes:
This update makes it clear when running the test-aux tool that the memory attribute tests are currently skipped as I have not implemented the page table functionality (if even possible). It does this by reporting skipped tests instead of quietly passing it.

## Description

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Compiles

## Integration Instructions

N/A

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/test_aux/src/efi.rs
+++ b/SeaPkg/Tools/GenSeaArtifacts/test_aux/src/efi.rs
@@ -12,8 +12,8 @@ use std::{
 
 use r_efi::efi::Status;
 
-const WIDTH: usize = 8usize * core::mem::size_of::<Status>();
-const WARNING_MASK: usize = 0x0 << (WIDTH - 8);
+const STATUS_WIDTH: usize = 8usize * core::mem::size_of::<Status>();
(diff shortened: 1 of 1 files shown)


kuqin12 (approved):

SeaPkg/Tools/GenSeaArtifacts/test_aux/src/efi.rs
```diff
@@ -12,6 +12,11 @@ use std::{
 
 use r_efi::efi::Status;
 
+const WIDTH: usize = 8usize * core::mem::size_of::<Status>();
+const WARNING_MASK: usize = 0x0 << (WIDTH - 8);
```
makubacki: nit: `STATUS_WIDTH` is more precise.

```suggestion
const STATUS_WIDTH: usize = 8usize * core::mem::size_of::<Status>();
const WARNING_MASK: usize = 0x0 << (STATUS_WIDTH - 8);
```

vineelko (approved):

PR #595, opened 2025-12-18
Add a dummy `AsmInvd` function to suffice the usage from secondary libraries

kuqin12:
## Description

The `CacheMaintenaceLib` is implementing its own functions based on a few `BaseLib` primitives.

Thus for any supervised standalone MM module that links cache maintenance library, the build could break.

This change bring back the function symbol but do not provide any implementation to resolve the build break but does not change the functionality of the original code.

- [ ] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on physical hardware platform and fixed the build.

## Integration Instructions

N/A

### DIFF
--- a/MmSupervisorPkg/Library/BaseLibSysCall/X64/Invd.c
+++ b/MmSupervisorPkg/Library/BaseLibSysCall/X64/Invd.c
@@ -9,6 +9,7 @@
 
 #include    <Uefi.h>
 #include    <Library/BaseLib.h>
+#include    <Library/DebugLib.h>
 
 /**
   Executes a INVD instruction.


MmSupervisorPkg/Library/BaseLibSysCall/X64/Invd.c
```diff
@@ -0,0 +1,33 @@
...
+
+**/
+
+/**
+  Microsoft Visual Studio 7.1 Function Prototypes for I/O Intrinsics.
+**/
+
+#include    <Uefi.h>
+#include    <Library/BaseLib.h>
+#include    <Library/SysCallLib.h>
```
kuqin12: no need.

kuqin12: Removed.

vineelko (approved):

PR #600, opened 2026-01-07
MmSupervisorPkg: Add MmCommunicate V3 support.

apop5:
## Description

Add EFI_MM_COMMUNICATE_HEADER_V3 support to Mm Supervisor

This is adding support for processing mm communicate v3 requests from the protocol.

v3 support is not being added to the supervisor communication channel. The supervisor channel, as can be viewed in MmDxeSupport.c, directly crafts a v2 communicate packet.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Testing Q35 with 202511. 
Would not boot previously, asserted when mm communicate v3 attempted to access variables. 
Booted after

## Integration Instructions
No integration necessary, should be backward compatible previous mm communicate protocols.

### DIFF
--- a/MmSupervisorPkg/Core/MmSupervisorCore.c
+++ b/MmSupervisorPkg/Core/MmSupervisorCore.c
@@ -516,7 +516,6 @@ MmEntryPoint (
 
         // Check if MM Communicate V3 is being used
         if (CompareGuid (&(CommunicateHeader->HeaderGuid), &gEfiMmCommunicateHeaderV3Guid)) {
-          DEBUG ((DEBUG_ERROR, "V3 Comm is being used\n"));
           CommunicateHeaderV3 = (EFI_MM_COMMUNICATE_HEADER_V3 *)CommunicateHeader;
           CommGuidOffset      = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER_V3, MessageGuid);
           CommHeaderSize      = sizeof (EFI_MM_COMMUNICATE_HEADER_V3);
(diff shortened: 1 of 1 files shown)


MmSupervisorPkg/Core/MmSupervisorCore.c
```diff
@@ -557,13 +571,21 @@ MmEntryPoint (
...
+        if (CompareGuid (&(CommunicateHeader->HeaderGuid), &gEfiMmCommunicateHeaderV3Guid)) {
+          CommunicateHeaderV3 = (EFI_MM_COMMUNICATE_HEADER_V3 *)CommunicateHeader;
+          CommGuidOffset      = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER_V3, MessageGuid);
+          CommHeaderSize      = sizeof (EFI_MM_COMMUNICATE_HEADER_V3);
+          BufferSize          = CommunicateHeaderV3->BufferSize;
+        } else {
+          CommGuidOffset = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, HeaderGuid);
+          CommHeaderSize = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data);
+          BufferSize     = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data) + CommunicateHeader->MessageLength;
+        }
```
kuqin12: I am wondering if we would want to support v3 for the supervisor protocol at all? We would need to update the caller side as well if that is the case.

MmSupervisorPkg/Core/MmSupervisorCore.c
```diff
@@ -511,37 +514,48 @@ MmEntryPoint (
...
+        BufferSize                                -= CommHeaderSize;
+        SupervisorToUserDataBuffer->UserBufferSize = BufferSize;
+
         Status = MmiManage (
-                   &CommunicateHeader->HeaderGuid,
+                   (EFI_GUID *)((UINT8 *)CommunicateHeader + CommGuidOffset),
                    NULL,
-                   CommunicateHeader->Data,
+                   (UINT8 *)CommunicateHeader + CommHeaderSize,
                    (UINTN *)&(SupervisorToUserDataBuffer->UserBufferSize)
```
kuqin12: I think we have reviewed this part together close enough :)

MmSupervisorPkg/Core/MmSupervisorCore.c
makubacki: After the call to `MmiManage()` on the supervisor path, it still does:

```c
BufferSize = BufferSize + OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data);
```

From a maintainability perspective, since other calculations were already updated on that path, using `BufferSize` that was previously set would be make usage consistent: `BufferSize = BufferSize + CommHeaderSize;`

PR #603, opened 2026-01-12
MmSupervisorPkg: Add back override tracking tags to MmSupervisorCore

apop5:
## Description

Override tags were removed in #600.

Add back the override tags to catch changes.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
CI

## Integration Instructions
No integration necessary.

### DIFF
--- a/MmSupervisorPkg/Core/MmSupervisorCore.inf
+++ b/MmSupervisorPkg/Core/MmSupervisorCore.inf
@@ -10,7 +10,15 @@
 #
 ##
 
+# release/202502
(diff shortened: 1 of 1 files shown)


makubacki:
@apop5, my understanding was that was intentional at the time. Was it and this is a decision to revert to having the tags again? Or was that unexpected?

apop5:
> @apop5, my understanding was that was intentional at the time. Was it and this is a decision to revert to having the tags again? Or was that unexpected?

it was intentional at the time for testing, but not for the final commit. It was an oversight to have it in the PR.

PR #636, opened 2026-04-07
Support reblock

kuqin12:
## Description

Current method of migrating MM accessible buffers is leaving the previously unblocked buffers unblocked, which will pose the risk of failing DRTM validation.

This change adds the implementation of `MmReblockMemoryRequest` newly introduced in basecore:
- The supervisor will support a new opcode that supports reblocking a previously unblocked region
- The unblock memory driver will support a new protocol interface to accept reblocking requests
- The newly added protocol interface will serve as the back end of `MmReblockMemoryRequest`
- The supervisor will stop accepting the reblocking request after ready to lock event

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested using QEMU Q35 and booted to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/MmSupervisorPkg/Core/Request/Request.h
+++ b/MmSupervisorPkg/Core/Request/Request.h
@@ -95,4 +95,25 @@ ProcessUpdateCommBufferRequest (
   IN MM_SUPERVISOR_COMM_UPDATE_BUFFER  *UpdateCommBuffer
   );
 
+/**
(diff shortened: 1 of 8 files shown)


kuqin12:
Ended up not needing it as we covered all the corner cases.

PR #655, opened 2026-05-20
Close supervisor communication channel at Exit Boot Services

### LINKED ISSUE #654: [Bug]: Block supervisor communication channel after EBS
### Is there an existing issue for this?

- [x] I have searched existing issues

### Current Behavior

This bug is created to track the change on shutting down the supervisor channel after the exit boot services event has occurred.

### Expected Behavior

Supervisor will not be able to be talked to after EBS. It was created to fetch the policy for FASR and testing purpose. We should be able to safely close it after EBS.

Note that we will not be revoking the access to the unblocked supervisor regions because that will mutate the memory policy, which is against the rule of memory policy will not change after being fetched.

### Steps To Reproduce

Just boot up the system and communicate with supervisor.

### Build Environment

```markdown
- OS(s): Windows 11
- Tool Chain(s): MSVC + CLANGPDB on x86
- Targets Impacted: DEBUG/RELEASE
```

### Version Information

```text
Top of main
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

kuqin12:
## Description

This change updates the Supervisor communication DXE agent to signal the exit boot services to supervisor in order to shut down the supervisor communication channel after the control is handed off to the OS.

The change also flips the communication buffer consumption order, which prioritize the user buffer serving first, then the supervisor buffer.

Resolves #654.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU Q35 and booted to Windows desktop with `MmExitBootServicesHandler` printed in the log.

## Integration Instructions

N/A

### DIFF
--- a/MmSupervisorPkg/Core/MmSupervisorCore.c
+++ b/MmSupervisorPkg/Core/MmSupervisorCore.c
@@ -447,11 +447,11 @@ MmReadyToLockHandler (
   This function will set a flag to indicate MM is at runtime, and the flag will be used to prevent
   any further MM communication through supervisor buffer.
 
-  @param  DispatchHandle  The unique handle assigned to this handler by MmiHandlerRegister().
-  @param  Context         Points to an optional handler context which was specified when the handler was registered.
-  @param  CommBuffer      A pointer to a collection of data in memory that will
-                          be conveyed from a non-MM environment into an MM environment.
-  @param  CommBufferSize  The size of the CommBuffer.
+  @param[in]      DispatchHandle  The unique handle assigned to this handler by MmiHandlerRegister().
+  @param[in]      Context         Points to an optional handler context which was specified when the handler was registered.
+  @param[in,out]  CommBuffer      A pointer to a collection of data in memory that will
+                                  be conveyed from a non-MM environment into an MM environment.
+  @param[in,out]  CommBufferSize  The size of the CommBuffer.
 
   @return Status Code
 
--- a/MmSupervisorPkg/Core/MmSupervisorCore.c
+++ b/MmSupervisorPkg/Core/MmSupervisorCore.c
@@ -108,10 +108,11 @@ EFI_MEMORY_DESCRIPTOR  mMmSupervisorAccessBuffer[MM_OPEN_BUFFER_CNT];
 // Table of MMI Handlers that are registered by the MM Core when it is initialized
 //
 MM_CORE_MMI_HANDLERS  mMmCoreMmiHandlers[] = {
-  { MmDriverDispatchHandler, &gMmSupervisorDriverDispatchGuid,  NULL, TRUE  },
-  { MmReadyToLockHandler,    &gEfiDxeMmReadyToLockProtocolGuid, NULL, TRUE  },
-  { MmSupvRequestHandler,    &gMmSupervisorRequestHandlerGuid,  NULL, FALSE },
-  { NULL,                    NULL,                              NULL, FALSE },
+  { MmDriverDispatchHandler,    &gMmSupervisorDriverDispatchGuid,  NULL, TRUE  },
+  { MmReadyToLockHandler,       &gEfiDxeMmReadyToLockProtocolGuid, NULL, TRUE  },
+  { MmSupvRequestHandler,       &gMmSupervisorRequestHandlerGuid,  NULL, FALSE },
+  { MmExitBootServicesHandler,  &gEfiEventExitBootServicesGuid,    NULL, FALSE },
+  { NULL,                    NULL,                                 NULL, FALSE },
 };
 
 EFI_SYSTEM_TABLE                  *mEfiSystemTable;
@@ -121,6 +122,7 @@ EFI_MM_DRIVER_ENTRY               *mMmCoreDriverEntry;
 MM_SUPV_USER_COMMON_BUFFER        *SupervisorToUserDataBuffer = NULL;
 BOOLEAN                           mMmReadyToLockDone          = FALSE;
 BOOLEAN                           mCoreInitializationComplete = FALSE;
+BOOLEAN                           mAtRuntime                  = FALSE;
 VOID                              *mInternalCommBufferCopy[MM_OPEN_BUFFER_CNT];
 SMM_SUPV_SECURE_POLICY_DATA_V1_0  *FirmwarePolicy = NULL;
 
@@ -441,6 +443,45 @@ MmReadyToLockHandler (
 }
 
 /**
+  Software MMI handler that should be triggered from non-MM environment upon DxeMmReadyToLock
+  event. This function unregisters the SUPERVISOR MMIs that are not required after Ready To Lock
+  event. Certain features, such as unblock memory regions, will not be available after this point.
+
+  Note: User ready to lock event will be notified prior to this supervisor ready to lock event.
+  This order is controlled in the corresponding DXE agent (IPL and/or DxeSupport driver).
+
+  @param  DispatchHandle  The unique handle assigned to this handler by MmiHandlerRegister().
+  @param  Context         Points to an optional handler context which was specified when the handler was registered.
+  @param  CommBuffer      A pointer to a collection of data in memory that will
+                          be conveyed from a non-MM environment into an MM environment.
+  @param  CommBufferSize  The size of the CommBuffer.
+
+  @return Status Code
+
+**/
+EFI_STATUS
+EFIAPI
+MmExitBootServicesHandler (
+  IN     EFI_HANDLE  DispatchHandle,
+  IN     CONST VOID  *Context         OPTIONAL,
+  IN OUT VOID        *CommBuffer      OPTIONAL,
+  IN OUT UINTN       *CommBufferSize  OPTIONAL
+  )
+{
+  DEBUG ((DEBUG_INFO, "%a\n", __func__));
+
+  if (mAtRuntime) {
+    DEBUG ((DEBUG_ERROR, "ExitBootServices event is signaled more than once??\n"));
+    ASSERT (FALSE);
+    return EFI_ALREADY_STARTED;
+  }
+
+  mAtRuntime = TRUE;
+
+  return EFI_SUCCESS;
+}
+
+/**
   Determine if two buffers overlap in memory.
 
   @param[in] Buff1  Pointer to first buffer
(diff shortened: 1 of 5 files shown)


MmSupervisorPkg/Core/MmSupervisorCore.c
```diff
@@ -440,6 +442,45 @@ MmReadyToLockHandler (
   return Status;
 }
 
+/**
+  Software MMI handler that should be triggered from non-MM environment upon DxeMmReadyToLock
```
makubacki: Need to update the function description for EBS (referring to Ready to Lock) and what it's doing here.

MmSupervisorPkg/Drivers/MmPeiLaunchers/MmDxeSupport.c
```diff
@@ -346,6 +364,56 @@ SmmIplReadyToLockEventNotify (
...
+
+  //
+  // Use Guid to initialize EFI_SMM_COMMUNICATE_HEADER structure
+  //
+  CopyGuid (&mCommunicateHeader->HeaderGuid, (EFI_GUID *)Context);
+
+  //
+  // Check to make sure the header GUID is the correct one for ExitBootServices.
+  //
+  if (!CompareGuid ((EFI_GUID *)Context, &gEfiEventExitBootServicesGuid)) {
```
makubacki, line 392: Did you mean to check this before the `CopyGuid()` above?

MmSupervisorPkg/Core/MmSupervisorCore.c
```diff
@@ -532,124 +573,124 @@ MmEntryPoint (
     CopyMem (&MmCommunicationUserBufferStatus, mMmCommUserMailboxBufferStatus, sizeof (*mMmCommUserMailboxBufferStatus));
   }
 
-  if (MmCommunicationSupvBufferStatus.IsCommBufferValid) {
+  if (MmCommunicationUserBufferStatus.IsCommBufferValid) {
```
makubacki, line 573: Afaict, this doesn't handle the `mAtRuntime && MmCommunicationSupvBufferStatus.IsCommBufferValid` case. I know it's very unlikely.

It looks like it might fall to async MMIs. I think you might want to explicitly clear `IsCommBufferValid` and return a status like `EFI_ACCESS_DENIED`.

MmSupervisorPkg/Core/MmSupervisorCore.c
makubacki: Description in the .h file needs to be updated too.

MmSupervisorPkg/Drivers/MmPeiLaunchers/MmDxeSupport.c
```diff
@@ -346,6 +364,56 @@ SmmIplReadyToLockEventNotify (
 
 // MM_SUPV: Update communicate buffer when entering DXE.
 
+/**
+  Event notification that is fired when the ExitBootServices event is signaled.
+
+  @param  Event                 The Event that is being processed, not used.
+  @param  Context               Event Context, not used.
```
makubacki:
```suggestion
  @param[in]  Event                 The Event that is being processed, not used.
  @param[in]  Context               Event Context, not used.
```

MmSupervisorPkg/Drivers/MmPeiLaunchers/MmDxeSupport.c
```diff
@@ -346,6 +364,56 @@ SmmIplReadyToLockEventNotify (
...
+  //
+  // Check to make sure the header GUID is the correct one for ExitBootServices.
+  //
+  if (!CompareGuid ((EFI_GUID *)Context, &gEfiEventExitBootServicesGuid)) {
+    DEBUG ((DEBUG_ERROR, "Unexpected event notification for %g\n", (EFI_GUID *)Context));
+    ASSERT (FALSE);
+    return;
+  }
+
+  mCommunicateHeader->MessageLength = 1;
```
makubacki, line 404: Could you please leave a comment explaining why `MessageLength` is set to `1` here?

MmSupervisorPkg/Drivers/MmPeiLaunchers/MmDxeSupport.c
```diff
@@ -103,6 +103,20 @@ SmmIplEndOfDxeEventNotify (
   IN VOID       *Context
   );
 
+/**
+  Event notification that is fired when the ExitBootServices event is signaled.
+
+  @param  Event                 The Event that is being processed, not used.
+  @param  Context               Event Context, not used.
```
makubacki:
```suggestion
  @param[in]  Event                 The Event that is being processed, not used.
  @param[in]  Context               Event Context, not used.
```

MmSupervisorPkg/Drivers/MmPeiLaunchers/MmDxeSupport.c
kuqin12: Yeah, I thought about this when I did it and ended up checking the final content. Can move this up to check against the incoming before copying.

MmSupervisorPkg/Core/MmSupervisorCore.c
kuqin12: Agreed. This is updated. Will push shortly.

MmSupervisorPkg/Core/MmSupervisorCore.c
kuqin12: This is updated

MmSupervisorPkg/Core/MmSupervisorCore.c
kuqin12: Also updated.

MmSupervisorPkg/Drivers/MmPeiLaunchers/MmDxeSupport.c
kuqin12: This is updated.

MmSupervisorPkg/Core/MmSupervisorCore.c
```diff
@@ -440,6 +442,42 @@ MmReadyToLockHandler (
...
+  IN OUT VOID        *CommBuffer      OPTIONAL,
+  IN OUT UINTN       *CommBufferSize  OPTIONAL
+  )
+{
+  DEBUG ((DEBUG_INFO, "%a\n", __func__));
+
+  if (mAtRuntime) {
+    DEBUG ((DEBUG_ERROR, "ExitBootServices event is signaled more than once??\n"));
+    ASSERT (FALSE);
+    return EFI_ALREADY_STARTED;
```
apop5: I don't think EFI_ALREADY_STARTED is a valid return code from a MmiHandler

https://uefi.org/specs/PI/1.9/V4_Management_Mode_System_Table.html#mmihandlerregister

Table 3.48 Status Codes Returned (EFI_MM_HANDLER_ENTRY_POINT)

EFI_SUCCESS should be returned if the MMI was handled, or EFI_WARN_INTERRUPT_SOURCE_QUIESCED

kuqin12: Yes, this is once again falling into the PI spec only allows 3 return status code in this case. Any suggestions?

MmSupervisorPkg/Core/MmSupervisorCore.c
```diff
@@ -440,6 +442,42 @@ MmReadyToLockHandler (
...
+/**
+  Software MMI handler that is called when ExitBootServices is triggered in non-MM environment.
+  This function will set a flag to indicate MM is at runtime, and the flag will be used to prevent
+  any further MM communication through supervisor buffer.
+
+  @param  DispatchHandle  The unique handle assigned to this handler by MmiHandlerRegister().
+  @param  Context         Points to an optional handler context which was specified when the handler was registered.
+  @param  CommBuffer      A pointer to a collection of data in memory that will
+                          be conveyed from a non-MM environment into an MM environment.
+  @param  CommBufferSize  The size of the CommBuffer.
```
makubacki: Sorry that I missed these in my first pass.

```suggestion
  @param[in]      DispatchHandle  The unique handle assigned to this handler by MmiHandlerRegister().
  @param[in]      Context         Points to an optional handler context which was specified when the handler was registered.
  @param[in,out]  CommBuffer      A pointer to a collection of data in memory that will
                                  be conveyed from a non-MM environment into an MM environment.
  @param[in,out]  CommBufferSize  The size of the CommBuffer.
```

MmSupervisorPkg/Core/MmSupervisorCore.h
```diff
@@ -802,6 +802,29 @@ MmReadyToLockHandler (
...
+/**
+  Software MMI handler that is called when ExitBootServices is triggered in non-MM environment.
+  This function will set a flag to indicate MM is at runtime, and the flag will be used to prevent
+  any further MM communication through supervisor buffer.
+
+  @param  DispatchHandle  The unique handle assigned to this handler by MmiHandlerRegister().
+  @param  Context         Points to an optional handler context which was specified when the handler was registered.
+  @param  CommBuffer      A pointer to a collection of data in memory that will
+                          be conveyed from a non-MM environment into an MM environment.
+  @param  CommBufferSize  The size of the CommBuffer.
```
makubacki:
```suggestion
  @param[in]      DispatchHandle  The unique handle assigned to this handler by MmiHandlerRegister().
  @param[in]      Context         Points to an optional handler context which was specified when the handler was registered.
  @param[in,out]  CommBuffer      A pointer to a collection of data in memory that will
                                  be conveyed from a non-MM environment into an MM environment.
  @param[in,out]  CommBufferSize  The size of the CommBuffer.
```

makubacki (approved):

MmSupervisorPkg/Core/MmSupervisorCore.h
kuqin12: Good call, thanks!

MmSupervisorPkg/Core/MmSupervisorCore.c
apop5: I think the warning plus EFI_SUCCESS should be fine, but it would stop processing any other handlers.

EFI_WARN_INTERRUPT_SOURCE_QUIESCED is the other options, it would say that other MMI handlers should still be processed, but given that it from the mm communicate buffer, there should really only be one handlers registered for it, so it is not like it really matters.

kuqin12: this going to be an EBS handler specific to the supervisor, so there really should not be other handlers. But warning + success seems acceptable. Will update.

kuqin12: Updated to warning only

apop5 (approved):

PR #664, opened 2026-07-01
[gen_aux] Allow the non-forward-reference symbols of zero length

kuqin12:
## Description

This issue was found when running this tool against the rust supervisor. The **NopLogger** pulled in as part of the logging crate has this field of 0 length.

This type of symbol are genuine Zero-Sized Type (ZST) and we should allow it. For other zero-sized symbol, we will keep the errors as before.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested against the rust supervisor and fixed the tool exception.

## Integration Instructions

N/A

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
+++ b/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
@@ -856,6 +856,11 @@ impl TypeInfo {
             if d.size != 0 {
                 return Ok(data);
             }
+            // A size-0 class that is a complete definition (not a forward
+            // reference) is a genuine zero-sized type. Return it as-is.
+            if !d.properties.forward_reference() {
+                return Ok(data);
+            }
             class_name = d.name.to_string().to_string();
         } else {
             return Ok(data);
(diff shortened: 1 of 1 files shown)


SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
makubacki:
```suggestion
// Return the item as-is if it is anything other than a class, or if it
// is a class that is not a zero-size forward reference.
```

makubacki:
```suggestion
// The type was a size-0 forward-reference class, so it should have a
// shadow class with the real information.
```

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
```diff
@@ -856,6 +856,11 @@ impl TypeInfo {
             if d.size != 0 {
                 return Ok(data);
             }
+            // A size-0 class that is a complete definition (not a forward
+            // reference) is a genuine zero-sized type. Return it as-is.
+            if !d.properties.forward_reference() {
+                return Ok(data);
```
makubacki, line 861: Not super familiar with the flow but a potential general issue with this type of change would be someone assuming the size returned is non-zero and using in places like a denominator where it might panic.

Is that a potential issue? Or did you rule it out?

kuqin12: Good call. Scanned through the module and added a guard against the usage in line 805.

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
kuqin12: Updated.

makubacki (approved):

Javagedes (approved):

PR #666, opened 2026-07-10
MmSupervisorPkg: Use SafeInt for Buffer Validation

apop5:
## Description

Use SafeIntLib for buffer size validation and manipulation.

- [ ] Impacts functionality?
- [X] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Booted Q35.

## Integration Instructions
No integration neccessary.

### DIFF
--- a/MmSupervisorPkg/Core/MmSupervisorCore.c
+++ b/MmSupervisorPkg/Core/MmSupervisorCore.c
@@ -588,7 +588,14 @@ MmEntryPoint (
     } else {
       CommGuidOffset = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, HeaderGuid);
       CommHeaderSize = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data);
-      BufferSize     = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data) + CommunicateHeader->MessageLength;
+      if (EFI_ERROR (SafeUint64Add (OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data), CommunicateHeader->MessageLength, &BufferSize))) {
+        DEBUG ((DEBUG_ERROR, "%a User comm header + MessageLength (0x%lx) overflows UINT64.\n", __func__, CommunicateHeader->MessageLength));
+        ASSERT (FALSE);
+        MmCommunicationUserBufferStatus.IsCommBufferValid = FALSE;
+        MmCommunicationUserBufferStatus.ReturnBufferSize  = 0;
+        MmCommunicationUserBufferStatus.ReturnStatus      = EFI_BAD_BUFFER_SIZE;
+        goto Cleanup;
+      }
     }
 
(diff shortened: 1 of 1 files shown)


kuqin12 (approved):

MmSupervisorPkg/Core/MmSupervisorCore.c
```diff
@@ -669,7 +699,14 @@ MmEntryPoint (
...
 
-    BufferSize -= CommHeaderSize;
+    if (EFI_ERROR (SafeUint64Sub (BufferSize, CommHeaderSize, &BufferSize))) {
+      DEBUG ((DEBUG_ERROR, "%a Supervisor BufferSize (0x%lx) is smaller than CommHeaderSize (0x%x), comm buffer is malformed.\n", __func__, BufferSize, CommHeaderSize));
+      ASSERT (FALSE);
+      MmCommunicationSupvBufferStatus.IsCommBufferValid = FALSE;
+      MmCommunicationSupvBufferStatus.ReturnBufferSize  = 0;
+      MmCommunicationSupvBufferStatus.ReturnStatus      = EFI_BAD_BUFFER_SIZE;
+      goto Cleanup;
+    }
```
makubacki: I think this was fine with normal subtraction: `BufferSize -= CommHeaderSize;`

`gEfiMmCommunicateHeaderV3Guid` is always rejected and would `goto Cleanup`:

---

```c
if (CompareGuid (&(CommunicateHeader->HeaderGuid), &gEfiMmCommunicateHeaderV3Guid)) {
  // ...
  goto Cleanup;
} else {
  CommHeaderSize = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data);
  SafeUint64Add (OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data), CommunicateHeader->MessageLength, &BufferSize);
  // (goto Cleanup on overflow)
}
```

So, to be here (what is currently Line 702), the `else` block was executed. Which now uses `SafeUint64Add()` for `BufferSize`'s assignment. If that failed, it would also `goto Cleanup`.

Since `CommunicateHeader->MessageLength` is unsigned (`UINT64`), it must be >= 0. So, `BufferSize` is guaranteed to be `CommHeaderSize` (which is `OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data)`) + `CommunicateHeader->MessageLength` (a whole number). In which case, `CommHeaderSize` is safe to subtract from it.

---

You could of course leave it, but I (and as noted by others in various reviews over time) prefer to only apply safe integer operations where required to limit them being over used and hiding cases where they're more relevant.

makubacki (approved):

MmSupervisorPkg/Core/MmSupervisorCore.c
apop5: Dropped the safeuint64sub in the latest push

kuqin12 (approved):

PR #674, opened 2026-08-19
[REBASE & FF] GenSeaArtifacts: Update script to allow different file name and padding bytes

kuqin12:
## Description

With the upcoming support of MM supervisor in rust, there are 2 issues with the current gen_aux script:

1. The current script hardcoded the supervisor file name to be `MmSupervisorCore`, which may not be true for binaries produced under rust projects.
2. With LLVM linker in place, the padding bytes beyond valid section size in the file is set to 0xCC, and not carried to memory during image loading. We will artificially inject 0s to the padding region only calculate the hash for SEA validation.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change includes unit test and also ran against the rust supervisor, the validation routine now passes.

## Integration Instructions

N/A

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
+++ b/SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
@@ -26,6 +26,7 @@ from edk2toollib.utility_functions import RunCmd, RunPythonScript
 
 HASH_ALGORITHM = "sha256"
 AUX_CONFIG_NAME = "AuxConfig.toml"
+MM_SUPERVISOR_CORE_NAME = "MmSupervisorCore"
 
 
 class GenSeaArtifacts(IUefiHelperPlugin):
@@ -44,7 +45,8 @@ class GenSeaArtifacts(IUefiHelperPlugin):
         mm_supervisor_build_dir: Path,
         mmi_file_path: Path,
         out_path: Path,
-        workspace=None
+        workspace=None,
+        supervisor_name=MM_SUPERVISOR_CORE_NAME,
     ):
         """Generates SEA artifacts.
 
@@ -63,13 +65,14 @@ class GenSeaArtifacts(IUefiHelperPlugin):
             mmi_file_path: Path to MmiEntrySea.bin file.
             out_path: Path to place all generated artifacts.
             workspace: Path to the workspace. If not provided, the current directory is used.
+            supervisor_name: Name of the supervisor. If not provided, a default name is used.
         """
         try:
             inc_file_path = out_path / "MmArtifacts.dsc.inc"
             temp_hash_dir = out_path / "temp_hash.bin"
             temp_out_dir = out_path / "temp_out.inc"
 
-            aux_path = generate_aux_file(aux_config_path, mm_supervisor_build_dir, scopes, out_path, workspace=workspace)
+            aux_path = generate_aux_file(aux_config_path, mm_supervisor_build_dir, scopes, out_path, workspace=workspace, supervisor_name=supervisor_name)
 
             # Copy the aux configuration file to the same directory as the aux file.
             shutil.copy2(aux_config_path, out_path / AUX_CONFIG_NAME)
@@ -108,7 +111,7 @@ class GenSeaArtifacts(IUefiHelperPlugin):
                 o.write("gEfiSeaPkgTokenSpaceGuid.PcdMmiEntryBinSize|0x%08X" % os.path.getsize(mmi_file_path))
 
             # MM supervisor core hash patching
-            mm_supv_file = mm_supervisor_build_dir / "MmSupervisorCore.efi"
+            mm_supv_file = mm_supervisor_build_dir / f"{supervisor_name}.efi"
             mm_supv_core_hash = calculate_loadable_image_hash(mm_supv_file)
             hex_bytes = bytes.fromhex(mm_supv_core_hash)
             with open(temp_hash_dir, 'wb') as f:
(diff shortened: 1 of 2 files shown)


SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
```diff
@@ -42,7 +43,8 @@ def generate_sea_includes(
         mm_supervisor_build_dir: Path,
         mmi_file_path: Path,
         out_path: Path,
-        workspace=None
+        workspace=None,
+        supervisor_name=None,
```
Javagedes: I would consider a const at the top

DEFAULT_SUPV_NAME="MmSupervisorCore"

Then in this function and the second function below do:

```python
def generate_sea_includes(
  ...
  supervisor_name=DEFAULT_SUPV_NAME
)
```

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
```diff
@@ -333,3 +344,53 @@ def calculate_file_hash(file: Path, offset: int = 0, length: int = -1):
...
+
+    data = bytearray(file.read_bytes())
+
+    if data[:2] != b"MZ":
+        raise ValueError(f"{file} does not start with a DOS header.")
+    pe_offset = struct.unpack_from("<I", data, 0x3C)[0]
+    if data[pe_offset:pe_offset + 4] != b"PE\0\0":
+        raise ValueError(f"{file} does not contain a PE signature.")
+
+    section_count = struct.unpack_from("<H", data, pe_offset + 6)[0]
```
Javagedes: I would use [pefile](https://pypi.org/project/pefile/) here instead of manually parsing the PECOFF file. We already have it installed as a part of `edk2-pytool-extensions`.

Javagedes (approved):

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
makubacki: I agree. At a minimum some basic checks like validating the file size seem to be missing.

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
```diff
@@ -333,3 +344,53 @@ def calculate_file_hash(file: Path, offset: int = 0, length: int = -1):
...
+        ValueError: The file is not a PE/COFF image, or a section extends past its end
+    """
+    if not file.exists():
+        raise FileNotFoundError(file)
+
+    data = bytearray(file.read_bytes())
+
+    if data[:2] != b"MZ":
+        raise ValueError(f"{file} does not start with a DOS header.")
+    pe_offset = struct.unpack_from("<I", data, 0x3C)[0]
```
makubacki: I think you should re-raise this exception to something documented in the function comment and handled in the caller (`generate_sea_includes()`). Right now, it just catches `FileNotFoundError` and `RuntimeError`.

You can raise how you'd like, but raising a `ValueError` (and handling it in the caller) is one way:

```python
#...
try:
  pe_offset = struct.unpack_from("<I", data, 0x3C)[0]
  #...
except struct.error as e:
    raise ValueError(f"{file} is not a well-formed PE/COFF image: {e}") from e
#...
```

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
```diff
@@ -292,8 +301,10 @@ def generate_aux_file(aux_config_path: Path, mm_supervisor_build_dir: Path, scop
 
     args = f"run --manifest-path {str(manifest_path)}"
     args += " --bin create-aux --"
-    args += f" --pdb {str(mm_supervisor_build_dir / 'MmSupervisorCore.pdb')}"
-    args += f" --efi {str(mm_supervisor_build_dir / 'MmSupervisorCore.efi')}"
+    if supervisor_name == None:
```
makubacki:
```suggestion
    if supervisor_name is None:
```

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
makubacki: Shouldn't this match the binary name?

makubacki (approved):

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
kuqin12: The difference was that this is an output file, so the script can dictate what the aux name would be. But I agree that we should be consistent here.

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
kuqin12: Good call. This is automatically updated to what you suggested after moving to use `pefile`.

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
kuqin12: With the default to `MmSupervisorCore`, this `None` check is removed.

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
kuqin12: This is updated.

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
kuqin12: Thanks for the suggestion. This is now updated.

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
kuqin12: Thanks, this is updated.

SeaPkg/Tools/GenSeaArtifacts/GenSeaArtifacts.py
makubacki: Right. I as a user would intuitively expect it to match the other name.

kuqin12:
Failing override check... Will fix in a separate PR.

kuqin12:
> Failing override check... Will fix in a separate PR.

Created here: https://github.com/microsoft/mu_feature_mm_supv/pull/676

vineelko (approved):

PR #677, opened 2026-08-24
[REBASE & FF] [gen_aux] Fixing field resolution for unions and added optional automatic descension logic

kuqin12:
## Description

This PR introduces 2 modifications on the field name resolution logic:

1. Added the support for union field.
2. Added a fallback logic to automatically descend to child fields if no exact match is found and a full blanket data layer is observed.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is tested with inline unit tests and verified on the rust-based supervisor binary.

## Integration Instructions

Before change:
```
[[rule]]
symbol = "mm_supervisor::LOGGER"
field = 'cell.data.value.value.inner.serial_port.blocking'
```

After change:
```
[[rule]]
symbol = "mm_supervisor::LOGGER"
field = 'cell.data.inner.serial_port.blocking'
```

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
+++ b/SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
@@ -20,6 +20,32 @@ use crate::{config, file, report};
 
 const POINTER_LENGTH: u64 = 8;
 
+/// A class or union record reduced to the parts this module needs, so one code path handles both.
+struct Aggregate<'t> {
+    name: pdb::RawString<'t>,
+    size: u64,
+    /// A class may have no field list; a union always has one.
+    fields: Option<TypeIndex>,
+    forward_reference: bool,
+}
+
+impl<'t> Aggregate<'t> {
+    fn new(data: &TypeData<'t>) -> Option<Self> {
+        let (name, size, fields, properties) = match data {
+            TypeData::Class(c) => (c.name, c.size, c.fields, c.properties),
+            TypeData::Union(u) => (u.name, u.size, Some(u.fields), u.properties),
+            _ => return None,
+        };
+
+        Some(Self {
+            name,
+            size,
+            fields,
+            forward_reference: properties.forward_reference(),
+        })
+    }
+}
+
 /// Additional context associated with a rule
 pub struct Context {
     pub name: String,
@@ -657,7 +683,7 @@ impl Symbol {
         }
     }
 
-    /// Returns the offset and size of a field in a class.
+    /// Returns the offset and size of a field in a class or union.
     fn find_field_offset_and_size(
         info: &TypeInformation,
         id: &TypeIndex,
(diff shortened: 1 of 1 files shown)


SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
```diff
@@ -868,34 +993,59 @@ impl TypeInfo {
...
 
-        // Return the item as-is if it is anything other than a class, or if it
-        // is a class that is not a zero-size forward reference.
-        let class_name;
-        if let TypeData::Class(d) = item {
-            if d.size != 0 {
-                return Ok(data);
+        // Anything that is not a zero-size forward reference is returned as-is. That includes a
+        // size-0 complete definition, which is a genuine zero-sized type rather than a stand-in.
+        let type_name = match item {
```
Javagedes: if `d` is the same type, then you can do this:

```rust
let type_name = match item {
    TypeData::Class(d) | TypeData::Union(d)
        if d.size == 0 && d.properties.forward_reference() =>
    {
        d.name.to_string().to_string()
    }
    _ => return Ok(data),
};
```

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
```diff
@@ -657,64 +657,185 @@ impl Symbol {
...
+        // Bound the walk so a malformed PDB cannot produce an unbounded modifier chain.
+        const MAX_MODIFIER_DEPTH: u32 = 32;
+
+        let mut index = index;
+        for _ in 0..MAX_MODIFIER_DEPTH {
+            let Ok(data) = TypeInfo::find_type(info, index).and_then(|item| Ok(item.parse()?))
+            else {
+                return false;
+            };
+            match data {
```
Javagedes, line 844: if `class` and `union` are the same underlying type then you can do this:

```rust
match data {
                // Qualifiers do not change the size of the underlying type.
                TypeData::Modifier(modifier) => index = modifier.underlying_type,
                TypeData::Class(d) | TypeData::Union(d) => {
                    return class.size == 0 && !class.properties.forward_reference()
                }
                _ => return false,
            }
```

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
```diff
@@ -868,34 +993,59 @@ impl TypeInfo {
...
+
+        // The type was a size-0 forward-reference class or union, so it should have a
+        // shadow definition with the real information.
+        let mut iter = info.iter();
+        let item = iter.find(|item| {
+            let item = item.parse()?;
+            if let Some(name) = item.name() {
+                if name.to_string() == type_name {
+                    match item {
+                        TypeData::Class(data) if data.size != 0 => return Ok(true),
```
Javagedes: same here... if they are the same underlying type :)

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
```diff
@@ -868,34 +993,59 @@ impl TypeInfo {
...
         let mut iter = info.iter();
         let item = iter.find(|item| {
             let item = item.parse()?;
             if let Some(name) = item.name() {
-                if name.to_string() == class_name {
-                    if let TypeData::Class(data) = item {
-                        if data.size != 0 {
-                            return Ok(true);
+                if name.to_string() == type_name {
+                    match item {
```
Javagedes: and here :)

At this point, if they are not the same underlying type it may be helpful to create a transformation that does make them the same underlying type to reduce these

Javagedes (approved):

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
kuqin12: Yes, good call here. I added an `Aggregate` to make the union and class to return the same value, so that the processor can handle them equivalently.

vineelko (approved):

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
kuqin12: And with the tranformer, this is not needed.

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
kuqin12: Not applicable any more.

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
kuqin12: Same here.

PR #678, opened 2026-08-26
[feat] gen_aux: support indexed nested array members

kuqin12:
## Description

Added an `array.field` to indicate the element name inside an array field. The original `field` is now used as the path to descend to the intended field in the structure.

The new `array.field` can be used in combination with the `array.index` to apply rules to specific fields inside an array.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested against the rust based supervisor and resolved all symbols inside the static variable.

## Integration Instructions

One could specify the following to validate the `entries` array inside IDT symbol, and ensures all the `offset_low` fields in each array element is not zero:

```
[[rule]]
symbol = "patina_internal_cpu::interrupts::x64::idt::IDT"
array.field = 'entries'
array.index = [0, 31]
field = 'offset_low'
validation.type = "non zero"
```

### DIFF
--- a/SeaPkg/Tools/GenSeaArtifacts/gen_aux/readme.md
+++ b/SeaPkg/Tools/GenSeaArtifacts/gen_aux/readme.md
@@ -72,13 +72,8 @@ remarks = 'Optional[String]'
 
 - `scope`: If specified, the rule is only applied when this scope is active. Otherwise it is always applied.
 - `symbol`: Determines the address and size for the rule
-- `field`: Updates the address and size to be that of the field, rather than the symbol itself. A symbol recovered
-from a linker map has no field names, because a map records only names and addresses, so its `field` is instead the
-index of one of the pieces the optimizer split it into, such as `'0'`. A `&dyn Trait` static, for example, is split
-into piece `0` for the data pointer and piece `1` for the vtable pointer.
-- `array.field`: Names an array member of the symbol, for when the array is nested inside the symbol rather than being
-the symbol itself. The rule then iterates that member, `array.index` refers to it, and `field` names a field of its
-element type. Without this the rule iterates the symbol's own elements, and `field` names a field of those.
+- `field`: Updates the address and size to be that of the field, rather than the symbol itself.
+- `array.field`: Names a member inside a structure that is an array. This could be used in conjunction with `field` to validate a "field" inside this array.
 - `array.index`: Only apply the rule to the specified index of the array, or inclusive range.
 - `array.sentinel`: Apply content rule to only the final rule such that its content must be all zeros.
 - `validation.type`: The type of validation to perform on this symbol. Different values may also require additional configuration
--- a/SeaPkg/Tools/GenSeaArtifacts/gen_aux/readme.md
+++ b/SeaPkg/Tools/GenSeaArtifacts/gen_aux/readme.md
@@ -61,6 +61,7 @@ comes with the following standard options:
 scope = 'Optional[String]'
 symbol = 'Required[String]'
 field = 'Optional[String]'
+array.field = 'Optional[String]'
 array.index = 'Optional[Int]'|'Optional[Array[Int;2]]'
 array.sentinel = 'Optional[Boolean]'
 validation.type = 'Required[String]'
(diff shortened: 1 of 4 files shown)


vineelko (approved):

vineelko (approved):

SeaPkg/Tools/GenSeaArtifacts/gen_aux/readme.md
```diff
@@ -72,7 +73,9 @@ remarks = 'Optional[String]'
 - `scope`: If specified, the rule is only applied when this scope is active. Otherwise it is always applied.
 - `symbol`: Determines the address and size for the rule
 - `field`: Updates the address and size to be that of the field, rather than the symbol itself.
-- `array.index`: Only apply the rule to the specified index of the array, or inclusive range
+- `array.field`: Names a member inside a structure that is an array. This could be used in conjunction with `field` to validate
+a "field" inside this array.
```
makubacki, line 77: This could be clarified a bit so there is no ambiguity around interpretation of "structure that is an array".


```suggestion
- `array.field`: Names a member of the symbol's structure that is itself an array, for when the array is nested rather than
  being the symbol itself.
```

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/config.rs
```diff
@@ -167,6 +167,13 @@ where
 #[derive(Debug, Serialize, Deserialize)]
```
makubacki, line 167: Have you considered deriving `Default` for `Array`? That should prevent having to explicitly add `field: None` in several places.

```suggestion
#[derive(Debug, Default, Serialize, Deserialize)]
```

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
```diff
@@ -184,9 +171,13 @@ impl<'a, S: Source<'a> + 'a> PdbMetadata<'a, S> {
                 [entry.offset as usize..(entry.offset + entry.size) as usize]
                 .to_vec();
 
+            // The index belongs to whichever array the rule iterates: the array member named by
+            // `array.field`, or the symbol itself.
             let mut name = rule.symbol.clone();
-            if element_count > 1 {
-                name += format!("[{}]", i).as_str();
+            match rule.array.as_ref().and_then(|array| array.field.as_deref()) {
```
makubacki, line 177: I may be misunderstanding, but doesn't `build_field_padding_entries()` also need to be updated to compensate for array indices being included in the name?

Javagedes (approved):
Personally I would prefer making this a breaking change then adding the confusion of field vs array.field being reversed. But I won't push for it any harder then this single comment :)

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/config.rs
kuqin12: Thanks. Good call. This is now updated.

SeaPkg/Tools/GenSeaArtifacts/gen_aux/src/metadata.rs
kuqin12: Thanks! Yes, you are right. The change missed this case. It added with a pairing unit test.

kuqin12:
> Personally I would prefer making this a breaking change then adding the confusion of field vs array.field being reversed. But I won't push for it any harder then this single comment :)

I was trying to avoid the breaking change because @vineelko will be buried in the signing pipeline :) But if eating the bullet here is preferred, we can make that happen.

Javagedes:
> > Personally I would prefer making this a breaking change then adding the confusion of field vs array.field being reversed. But I won't push for it any harder then this single comment :)
> 
> 
> 
> I was trying to avoid the breaking change because @vineelko will be buried in the signing pipeline :) But if eating the bullet here is preferred, we can make that happen.

I  defer to @vineelko then, as I completely understand the rough workload he already has:) we can always create an issue and change it once everything else has been handled :)

vineelko:
I think that's a good call. I agree with Joey's point, it does create some confusion. 

Since this is mainly introduced to better handle Rust symbols (for the Rust MM supervisor), should we retarget this PR to a feature branch with the semantics reversed and update the 3 rule files appropriately? In the meantime, we can get the final 2511 release going.

kuqin12:
> I think that's a good call. I agree with Joey's point, it does create some confusion.
> 
> Since this is mainly introduced to better handle Rust symbols (for the Rust MM supervisor), should we retarget this PR to a feature branch with the semantics reversed and update the 3 rule files appropriately? In the meantime, we can get the final 2511 release going.

Sure, we can table this till the final release is done. But I would hope to get this into the mainline directly to avoid mushing the policy changes in the other platforms together with the rust supv absorption.

I will make the change here to reverse the semantics and get it tested on the side.

kuqin12:
> > I think that's a good call. I agree with Joey's point, it does create some confusion.
> > Since this is mainly introduced to better handle Rust symbols (for the Rust MM supervisor), should we retarget this PR to a feature branch with the semantics reversed and update the 3 rule files appropriately? In the meantime, we can get the final 2511 release going.
> 
> Sure, we can table this till the final release is done. But I would hope to get this into the mainline directly to avoid mushing the policy changes in the other platforms together with the rust supv absorption.
> 
> I will make the change here to reverse the semantics and get it tested on the side.

The breaking change is made. Verified the test binary to reach DRTM + SMM level 30.

makubacki (approved):
Approving for the earlier feedback I left that's been resolved.

PR #684, opened 2026-08-31
[SEA] Add PCD to report all image validation violations before dumping the image

vineelko:
## Description

`PeCoffImageDiffValidation` stopped at the first failing entry, which requires
revising a new sea package and a new boot cycle to discover each subsequent
violation. This is especially tedious to do for memory attribute failures (which
cannot be validated by offline tool).  Accumulate the failures instead: log
every violating entry with its index, offset and rule type, keep walking the
entry list, and emit the MSEG info and image hex dump once at the end. Parsing
still stops early on an unknown rule type or a structural bounds failure since
the next entry location cannot be computed.

Based on the feedback, Added a `PcdImageValidationFailFast` to make this behavior 
an opt in for Debug builds by the consumer. For release builds and the default being
TRUE for the PCD, the first error will bailout from further validations.

Other minor typo related to `SAFE_DEBUG` macro where it is not defined for the `#else` case. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated on hardware platform

## Integration Instructions

NA

### DIFF
--- a/SeaPkg/Core/Init/StmInit.h
+++ b/SeaPkg/Core/Init/StmInit.h
@@ -41,7 +41,7 @@ extern SEA_HOST_CONTEXT_COMMON  mHostContextCommon;
       }                            \
     } while (FALSE)
 #else
-#define DEBUG(Expression)
+#define SAFE_DEBUG(Expression)
 #endif
 
 /**
--- a/SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
+++ b/SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
@@ -816,6 +816,8 @@ PeCoffImageDiffValidation (
   IMAGE_VALIDATION_ENTRY_HEADER  *NextImageValidationEntryHdr;
   UINTN                          Index;
   EFI_STATUS                     Status;
+  EFI_STATUS                     FinalStatus;
+  UINTN                          ViolationCount;
   EFI_PHYSICAL_ADDRESS           MsegBase;
   UINTN                          MsegSize;
   IMAGE_VALIDATION_MEM_ATTR      MsegMemAttr;
@@ -863,18 +865,25 @@ PeCoffImageDiffValidation (
     return Status;
   }
 
+  FinalStatus              = EFI_SUCCESS;
+  ViolationCount           = 0;
   ImageValidationEntryHdr  = (IMAGE_VALIDATION_ENTRY_HEADER *)((UINTN)ImageValidationHdr + ImageValidationHdr->OffsetToFirstEntry);
   OriginalImageLoadAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)(UINT8 *)OriginalImageBaseAddress;
   for (Index = 0; Index < ImageValidationHdr->EntryCount; Index++) {
     // TODO: Safe integer arithmetic
     if ((UINT8 *)(ImageValidationEntryHdr) >= ((UINT8 *)ImageValidationHdr + ImageValidationHdr->Size)) {
       DEBUG ((DEBUG_ERROR, "%a: Current header 0x%p exceeds the reference data limit 0x%x\n", __func__, ImageValidationEntryHdr, (UINT8 *)ImageValidationHdr + ImageValidationHdr->Size));
-      return EFI_COMPROMISED_DATA;
+      ViolationCount++;
+      FinalStatus = EFI_COMPROMISED_DATA;
+      // The entry stream cannot be parsed any further.
+      break;
     }
 
     if (ImageValidationEntryHdr->Offset + ImageValidationEntryHdr->Size > TargetImageSize) {
       DEBUG ((DEBUG_ERROR, "%a: Current entry range 0x%x exceeds target image limit 0x%x\n", __func__, ImageValidationEntryHdr->Offset + ImageValidationEntryHdr->Size, TargetImageSize));
-      return EFI_INVALID_PARAMETER;
+      ViolationCount++;
+      FinalStatus = EFI_INVALID_PARAMETER;
+      break;
     }
 
     // Ensure this entry's default value does not overflow the Auxiliary file buffer.
@@ -887,16 +896,10 @@ PeCoffImageDiffValidation (
         ImageValidationEntryHdr->OffsetToDefault + ImageValidationEntryHdr->Size,
         ImageValidationHdr->Size
         ));
-      return EFI_COMPROMISED_DATA;
+      ViolationCount++;
+      FinalStatus = EFI_COMPROMISED_DATA;
+      break;
     }
-
-    DEBUG ((
-      DEBUG_INFO,
-      "%a: Evaluating symbol at offset: 0x%x in the target image with rule type: 0x%x\n",
-      __func__,
-      ImageValidationEntryHdr->Offset,
-      ImageValidationEntryHdr->ValidationType
-      ));
 
     // All validation has been updated to reference the original image.  PeCoffLoaderRevertRelocateImage will
     // touch up various parts of the image that will include some pointers causing parts of the TargetImage to
@@ -928,7 +931,9 @@ PeCoffImageDiffValidation (
         break;
       default:
         Status = EFI_INVALID_PARAMETER;
-        // Does not support unknown validation type
+        // Does not support unknown validation type, the size of this entry is unknown so the
+        // location of the next entry cannot be computed.
+        NextImageValidationEntryHdr = NULL;
         DEBUG ((
           DEBUG_ERROR,
           "%a: Entry validation type not supported 0x%x\n",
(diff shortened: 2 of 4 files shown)


SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
```diff
@@ -816,6 +816,8 @@ PeCoffImageDiffValidation (
   IMAGE_VALIDATION_ENTRY_HEADER  *NextImageValidationEntryHdr;
```
Javagedes, line 817: This change is useful for DEBUG purposes but does not seem like it should exist in RELEASE. In my opinion, it seems unnecessarily dangerous to hit an error / failure case but continue verifying all rules. It also is not necessary in my opinion. While we will only report the first failure in the logs, we specifically dump the image we are validating, which can be used by the offline validation tool provided by this project to view all test failures at once. If some part of that flow lacks the data you need, then in my opinion we should be fixing it in the offline validation tool.

vineelko: The only problem with offline tool validation is validation of the memory attribute failures. Due to a recent bug fix https://github.com/microsoft/mu_feature_mm_supv/pull/680, there are some attribute related validation failures. It seemed very tedious to revise the SEA package for each failure and find the next one. Updated the PR description for more clarity. Feel free to let me know if you think otherwise.

Javagedes: This does definitely make sense. That is one shortfall we have. The main reason for that is that we did not have a good way to walk page tables at the time, so I did not add that functionality to the offline validation tool. It may be entirely possible to add that functionality using https://github.com/OpenDevicePartnership/patina-paging now.

Even if we don't want to go that route, this change really does not sit well with me always being available. If there is a way to only use it in debug builds, or via some other setting, that feels better. Not sure if others have opinions.

Javagedes: @makubacki @kuqin12 thoughts on this? Am I being overly cautious?

vineelko: Just to add, the last commit achieves that requested functionality by bailing out on the first violation based on `MDEPKG_NDEBUG`. But need others feedback if that is the right direction.

makubacki: I agree that we should "fail fast" in production scenarios. At that point, it doesn't matter if more rules are violated, as opposed to minimizing potential obstacles to get onto a known exit path.

`MDEPKG_NDEBUG` should not really be used for this. That's for debug code optimization and this is important enough it should be set clearly.

I'd add a direct config option for this that cannot be mixed up with other use cases, like a PCD. Default it to safe behavior (bailing on the first rule failure). Then, override it in the DSC on DEBUG builds. That's directly auditable in build reports, still works in C file conditions like this, and is obvious what's happening.

vineelko: Thanks all for the feedback, will address that.

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
```diff
@@ -939,25 +944,49 @@ PeCoffImageDiffValidation (
...
+
+      DEBUG ((
+        DEBUG_ERROR,
+        "Validation Error #%d! Entry index: 0x%x offset: 0x%x rule type: 0x%x - %r\n",
+        ViolationCount,
+        Index,
+        ImageValidationEntryHdr->Offset,
+        ImageValidationEntryHdr->ValidationType,
+        Status
+        ));
```
makubacki, line 978: Since `UINTN` will be 64-bit everywhere this is used, the print specifier should be `%Ld`. You could further cast the value to `UINT64` so it's guaranteed to match the integer width against either a 32-bit or 64-bit target, but I don't think it's really necessary here.

```suggestion
      DEBUG ((
        DEBUG_ERROR,
        "Validation Error #%Ld! Entry index: 0x%x offset: 0x%x rule type: 0x%x - %r\n",
        ViolationCount,
        Index,
        ImageValidationEntryHdr->Offset,
        ImageValidationEntryHdr->ValidationType,
        Status
        ));
```

SeaPkg/Core/Init/StmInit.h
```diff
@@ -41,7 +41,7 @@ extern SEA_HOST_CONTEXT_COMMON  mHostContextCommon;
       }                            \
     } while (FALSE)
 #else
-#define DEBUG(Expression)
+#define SAFE_DEBUG(Expression)
```
makubacki, line 44: Can you please call out this change in the PR description?

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
```diff
@@ -887,17 +896,11 @@ PeCoffImageDiffValidation (
         ImageValidationEntryHdr->OffsetToDefault + ImageValidationEntryHdr->Size,
         ImageValidationHdr->Size
         ));
-      return EFI_COMPROMISED_DATA;
+      ViolationCount++;
+      FinalStatus = EFI_COMPROMISED_DATA;
+      break;
     }
 
-    DEBUG ((
```
makubacki, line 893: Did you remove this message because it was too noisy? It might be worth keeping even if at a different debug level.

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
```diff
@@ -863,18 +865,25 @@ PeCoffImageDiffValidation (
...
+  ViolationCount           = 0;
   ImageValidationEntryHdr  = (IMAGE_VALIDATION_ENTRY_HEADER *)((UINTN)ImageValidationHdr + ImageValidationHdr->OffsetToFirstEntry);
   OriginalImageLoadAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)(UINT8 *)OriginalImageBaseAddress;
   for (Index = 0; Index < ImageValidationHdr->EntryCount; Index++) {
     // TODO: Safe integer arithmetic
     if ((UINT8 *)(ImageValidationEntryHdr) >= ((UINT8 *)ImageValidationHdr + ImageValidationHdr->Size)) {
       DEBUG ((DEBUG_ERROR, "%a: Current header 0x%p exceeds the reference data limit 0x%x\n", __func__, ImageValidationEntryHdr, (UINT8 *)ImageValidationHdr + ImageValidationHdr->Size));
-      return EFI_COMPROMISED_DATA;
+      ViolationCount++;
+      FinalStatus = EFI_COMPROMISED_DATA;
```
makubacki: Are these `FinalStatus` assignments meant to update the status if it was already set to an error?

For example, the code out of the switch block only sets `FinalStatus` if it not already set to an error. That would lead me to believe that the idea is that the actual status returned from the function (which is `FinalStatus`) is meant to be the first error detected. But these if statements overwrite it, so it seems to be inconsistent as to whether the error returned is the first or last error found.

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
vineelko: Fixed it.

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
vineelko: Yaa it was very noisy when I tested it and also felt redundant. In the latest iteration I converted it to `DEBUG_VERBOSE`

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
vineelko: I fixed the three if conditions and also changed the variable to `FirstErrorStatus` to make it explicit. Thanks.

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
vineelko: Added `PcdImageValidationFailFast`(Defaulted to TRUE) to break out of the validation loop.

SeaPkg/Library/BasePeCoffLibNegative/BasePeCoffLibNegative.c
makubacki: Thanks. It's been a while, but I seem to recall it being useful context to have.

makubacki (approved):
I'll leave it up to others as to whether they still have concerns with the PCD defaulted to safe behavior. I think that is reasonable to allow this to work and can be checked using other mechanisms (build time report checks, runtime checks, etc.) if needed.

kuqin12 (approved):

PR #687, opened 2026-09-02
[Rebase & FF] MmSupervisorPkg: Update override hash from latest mu_basecore release/202511

vineelko:
## Description

MmSupervisorCore: Update override hash from latest mu_basecore release/202511

Upstream StanadloneMmPkg changes:
```
749910060f48 StandaloneMmPkg Core: Check PE section finding status
04afff14ebbd StandaloneMmPkg Core: Treat no DEPEX as TRUE
12f8ff0fac4d StandaloneMmPkg/Core: Return when processing malformed DEPEX [Trivial]
```
Additional fix for 1, due to 2 with a NULL Depex now reachable,
MmAddToDriverList no longer calls AllocateCopyPool unconditionally.

Track tags for StandaloneMmCore.inf are refreshed to the release/202511.
This is needed for CI build to pass on main and https://github.com/microsoft/mu_feature_mm_supv/pull/684. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted on physical platform. Level 30 validation needs few rules related change in the sea releases.

## Integration Instructions

NA

### DIFF
--- a/MmSupervisorPkg/Core/Dispatcher/Dependency.c
+++ b/MmSupervisorPkg/Core/Dispatcher/Dependency.c
@@ -182,12 +182,10 @@ MmIsSchedulable (
 
   if (DriverEntry->Depex == NULL) {
     //
-    // A NULL Depex means that the MM driver is not built correctly.
-    // All MM drivers must have a valid depex expression.
+    // If there is no DEPEX, assume the module can be executed
     //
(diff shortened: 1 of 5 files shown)


apop5 (approved):

makubacki:
I think the bug fix is different enough from the (already needed) override hash update it should go in a separate commit.

makubacki (approved):

PR #692, opened 2026-09-22
[feature/supv] Moved the supervisor initializer into its own module

kuqin12:
## Description

This change moves the supervisor initializer specific files into its own module, rather than reusing the existing contract. So that the current supervisor is intact.

The hob creation routines are refactored into their own files and routines to allow cleaner readability.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is tested on QEMU Q35 and booted to Windows desktop.

## Integration Instructions

N/A


MmSupervisorPkg/Core/Hob/SupvInitHobs.c
kuqin12, line 1: This is the majority of the functional differences from this PR.

makubacki: I'm interested in reviewing the PR. Can you please call out in a bit more detail where the main functional changes are and what specifically moved around (source and destination locations)?

kuqin12: This called out file is the main change in this PR. It is intended to produce the initialization data for the rust supervisor to consume.

The MmSupervisorInit.c is the consumer of these interfaces and create hobs during the initialization process.

The original branch had the abstraction pattern, where the common file hosts a few callouts to supervisor/initializer specific implementations. But I think that will cause too much unnecessary perturbation for the mainline. So in this PR, I reverted the module to the original MM supervisor code and moved the initializer specific files into `*Init.c` files.

kuqin12:
Let's merge this first: https://github.com/microsoft/mu_feature_mm_supv/pull/696

kuqin12:
> Let's merge this first: #696

Done

